PR #11, opened 2020-05-09
Render Trait for Events

### LINKED ISSUE #2: Print out more event types.
The [RoomBuffer](https://github.com/poljar/weechat-matrix-rs/blob/master/src/room_buffer.rs) struct receives all the incoming events. It currently only prints out and handles properly a couple of events.

The event formatting should probably go into a separate file and should consist of pure methods. Something in the spirit of the [message_renderer](https://github.com/poljar/weechat-matrix/blob/master/matrix/message_renderer.py#L26) file from the python script.

### LINKED ISSUE #4: Users in the nicklist should use display names.
Currently the users in the nicklist are using the full mxid.

For this the [Room](https://github.com/matrix-org/matrix-rust-sdk/blob/master/src/base_client.rs#L59) struct in the rust-sdk would need to be enhanced to all the member changing events and provide a way to calculate the current display name of a user.

vbrandl:
This PR adds a `RenderEvent` trait that allows rendering events in the weechat UI (closes #2)

It also changed the nightly toolchain to `nightly-2020-04-06` which is the next toolchain that has both `clippy` and `fmt` available.

`calculate_user_name` was implemented to get the display name for a user in a room (closes #4)


vbrandl:
@poljar how do you test this project? Just load the plugin in weechat and join an unencrypted room?

poljar:
I have a local synapse setup, and yeah run it in Weechat is the only way to test it. Don't know if we can mock some Weechat methods somehow.

src/render.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn render(&self, displayname: &str) -> String {
+        let operation = match self.content.membership {
+            MembershipState::Join => "joined",
+            MembershipState::Leave => "left",
+            MembershipState::Ban => "banned",
+            MembershipState::Invite => "invited",
+            MembershipState::Knock => "knocked on", // TODO
+        };
+        format!(
+            "{} ({}) has {} the room",
```
poljar, line 37: We'll want to put some color into these, e.g. the delimiters. Another thing those messages have a prefix usually.

You can use `Weechat::color("chat_delimiters")` and `Weechat::prefix("join")` to get them.

src/render.rs
```diff
@@ -0,0 +1,140 @@
...
+impl RenderableEvent for EncryptedEvent {
+    // TODO: this is not implemented yet
+    fn render(&self, displayname: &str) -> String {
+        format!("{}\t{}", displayname, "Unable to decrypt message")
+    }
+}
+
+impl RenderableEvent for MemberEvent {
+    fn render(&self, displayname: &str) -> String {
+        let operation = match self.content.membership {
```
poljar, line 29: This isn't quite right, a `MembershipState::Join` will be present if a user joins or if a user changes the display name. You'll need to look at the `prev_content` and `content` to figure out the difference.

Luckily Ruma already has a convenience method for this, you can use [membership_change()](https://docs.rs/ruma-events/0.21.1/ruma_events/room/member/struct.MemberEvent.html#method.membership_change) on the event to get the correct change.

src/render.rs
```diff
@@ -0,0 +1,140 @@
...
+        };
+        format!(
+            "{} ({}) has {} the room",
+            displayname, self.state_key, operation
+        )
+    }
+}
+
+impl RenderableEvent for MessageEvent {
+    fn render(&self, displayname: &str) -> String {
```
poljar, line 44: I wonder if we should return the message and the tags here. (tags are metadata that are attached to the event, you can see them with `/debug tags`). For example we'll always want the user_id of the sender, the event_id,  perhaps some other stuff in the tags.

src/room_buffer.rs
```diff
@@ -165,6 +163,19 @@ impl RoomBuffer {
         self.weechat_buffer().set_name(&name)
     }
 
+    /// Calculate the display name for a room member from its UserId. If no member with that ID is in
+    /// the room, the string representation of the ID will be returned.
+    fn calculate_user_name(&self, user_id: &UserId) -> String {
```
poljar, line 168: This will need to implement display name disambiguation, e.g. when two users share the same display name we'll want to show the user id inside of brackets following the display name. Relevant spec entry can be found [here](https://matrix.org/docs/spec/client_server/latest#calculating-the-display-name-for-a-user).

src/render.rs
vbrandl: I'm new to writing weechat plugins. Both `color` and `prefix` return a `&str` but how do I use the return value? Is it used like [this](https://github.com/poljar/weechat-matrix-rs/pull/11/commits/519bbd2561c986702861a369b347bfe026786486)?

src/render.rs
vbrandl: Like this? e2d8e966d74a0dfa0bce6e40df1db18ca02bbebb

src/render.rs
poljar: Correct, you just join it into the final string. More examples of this can be found in the [Python script repo](https://github.com/poljar/weechat-matrix/blob/master/matrix/buffer.py#L721).

src/render.rs
poljar: Yes.

src/room_buffer.rs
vbrandl: Should be fixed by 9efea35af099e139bce0957701a252a845bcb7a5

src/render.rs
vbrandl: I went with adding `fn sender(&self) -> &UserId;` to the `RenderableEvent` trait. If needed, we can also add `fn event_id(&self) -> &EventId;`.

src/render.rs
vbrandl: I inserted some colors, but I'm not sure which prefix to use for regular messages.

Also currently we don't have per-user colors yet, so I just use the same color for each user

poljar: Regular messages don't have a prefix, at least not one that you'll fetch with `Weechat::prefix()`.

Colors for nicks can be fetched with `Weechat::info_get("nick_color_name", user_id)`, at least I think we want to use the user_id, so the color doesn't change when the display name changes.

Those will need to be stored somewhere like it's done [here](https://github.com/poljar/weechat-matrix/blob/master/matrix/buffer.py#L152).

But we can leave nick colors for another PR. They aren't required to get this merged.

src/room_buffer.rs
poljar: That's going to be a bit expensive to calculate every time and it should live in the rust-sdk but it's good enough for now.

src/render.rs
vbrandl: Then let's move per-user colors to another issue/PR if you are OK with this.

src/room_buffer.rs
vbrandl: Is it really? Getting from `HashMap` is `O(1)` and then `O(n)` for counting in the member list.
That's not the best but should perform OK.
Otherwise I think we need to hold a map from user ID to display name and update that on each join, quit, ...

poljar: Yeah the counting is the problem, it's done on every received message for every room. Though it'll be a much smaller problem in Rust than it would have been in Python since we can yield back control to Weechat between every event/message.

A map that gets updated on a join/quit sounds good. Joins/quits aren't received as often as normal message events. We can leave it as is but a PR to the matrix-sdk with a better algorithm would be very much appreciated.

src/render.rs
```diff
@@ -0,0 +1,236 @@
...
+            KickedAndBanned => "was kicked and banned",
+            InvitationRejected => "rejected the invitation",
+            InvitationRevoked => "had the invitation revoked",
+            ProfileChanged => "changed the display name or avatar",
+            NotImplemented => "performed an unimplemented operation",
+        };
+
+        let (prefix, color_action) = match change_op {
+            Joined | Invited => ("join", "green"),
+            _ => ("quit", "red"),
```
poljar, line 68: Nick changes have the network prefix on IRC, we should probably do the same.

src/render.rs
```diff
@@ -0,0 +1,236 @@
...
+            // TODO: formatting for inline markdown and so on
+            Text(t) => format!(
+                "{color_user}{}{color_reset}\t{}",
+                displayname,
+                t.resolve_body(),
+                color_user = color_user,
+                color_reset = color_reset
+            ),
+            Emote(e) => format!(
+                "{color_user}{}{color_reset}\t{}",
```
poljar, line 105: Emotes should be displayed as IRC `/me` messages, those use the `action` prefix.

src/render.rs
```diff
@@ -0,0 +1,236 @@
...
+            ),
+            Location(l) => format!(
+                "{color_user}{}{color_reset}\t{}: {}",
+                displayname,
+                l.body,
+                l.geo_uri,
+                color_user = color_user,
+                color_reset = color_reset
+            ),
+            Notice(n) => format!(
```
poljar, line 143: Notices should be rendered as IRC style notices. Like [here](https://github.com/poljar/weechat-matrix/blob/master/matrix/buffer.py#L601) or in the IRC plugin.

src/render.rs
```diff
@@ -0,0 +1,236 @@
...
+            ),
+            Video(v) => format!(
+                "{color_user}{}{color_reset}\t{}: {}",
+                displayname,
+                v.body,
+                v.resolve_url(),
+                color_user = color_user,
+                color_reset = color_reset
+            ),
+            ServerNotice(sn) => {
```
poljar, line 158: Should probably be rendered similarly to the normal notice.

src/render.rs
```diff
@@ -0,0 +1,236 @@
...
+    fn render(&self, displayname: &str) -> String;
+
+    /// Get the sender for this event this can be used to get the displayname.
+    fn sender(&self) -> &UserId;
+}
+
+impl RenderableEvent for EncryptedEvent {
+    // TODO: this is not implemented yet
+    fn render(&self, displayname: &str) -> String {
+        let color_user = Weechat::color("green"); // TODO: get per-user color
```
poljar, line 30: Since we're going to use the user_id to get the color we probably can get away without storing the color so a simple `Weechat::info_get("nick_color_name", user_id)` will suffice here, but we'll need to put some power level indicator before the display name as well.

For example if someone has a power level of 50 or above we will want to represent this using the `@` symbol like under IRC for ops. That info will be present in the room member struct. So we will probably need to pass the whole struct to the render method instead of just the display name.

But we can leave this for another PR.

src/render.rs
dkasak: I've handled this, but nick changes are still handled as joins for some reason. I'll take a look at it tomorrow.

dkasak: Fixed.

PR #13, opened 2020-07-21
Implement listfull command

lanker:


lanker:
OK. Thanks for merging!

PR #16, opened 2020-09-06
render: Create a Render trait using associated types and constants.

poljar:
This adds a render trait back, the trait now allows adding custom additional context that might be needed to render the event to be passed to the render method.

It also naturally scopes the event specific tags in the associated constant of the trait.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -1,4 +1,3 @@
-use crate::room_buffer::WeechatRoomMember;
 use matrix_sdk::{
     events::{
         room::{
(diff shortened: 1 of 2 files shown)


src/render.rs
```diff
@@ -15,169 +14,217 @@ use matrix_sdk::{
...
-                        color_delim = Weechat::color("chat_delimiters"),
-                        color_reset = Weechat::color("reset"))
-            }
+/// The rendered version of an event.
+pub struct RenderedEvent {
+    /// The UNIX timestamp of the event.
+    pub message_timestamp: u64,
+    /// The tags of the event,
+    pub tags: Vec<String>,
+    // TODO we need to allow multiple message/tags pairs here.
```
dkasak, line 28: Is there a reason (other than extra work) not to make messages/tags a `Vec<(String, Vec<String>)>` immediately? So we don't have to fiddle with it later on.

src/render.rs
```diff
@@ -15,169 +14,217 @@ use matrix_sdk::{
...
+}
+
+/// Trait allowing events to be rendered for Weechat.
+pub trait Render {
+    /// The event specific tags that should be attached to the rendered event.
+    const TAGS: &'static [&'static str];
+
+    /// Some events might need additional context to be rendered, for example if
+    /// we want to display the sender, we don't want to display the mxid,
+    /// instead we want to display the disambiguated display name.
```
dkasak, line 45:
```suggestion
    /// Some events might need additional context to be rendered. For example,
    /// instead of displaying the MXID for the sender, we might want to display
    /// the disambiguated display name, which isn't available in the event.
```

src/render.rs
poljar: No reason besides the extra work, though there's a bit more figuring out to do before I would change the print methods that take the rendered message to handle multiple message/tag pairs.

We would like to render the content separately but the content obviously doesn't have a timestamp nor sender. So the content needs to be renderable without a prefix (sender/timestamp in the case of normal messages).

How this should be done remains to be seen, two traits, some special casing or two trait methods...

dkasak:
To record my thoughts here too: this generally looks tidy and seems that it would work. Each event type might require some additional context of type `C`, which is an associated type, so the `C` varies with the concrete `render` implementation.

I do have a nagging feeling that I don't like using traits in this context, but since I don't have anything concrete to object to and it does look neat, I say go for it.

PR #38, opened 2021-04-04
convert mxc to http(s) or emxc

### LINKED ISSUE #36: Shows http links to file instead of mxc links
Currently when someone share a file, the displayed link is the mxc one, which is not directly usable. It should be translated to the http link with the homeserver url.

Austin-Ray:
Converts mxc to http(s) or emxc like the `weechat-matrix` python script does, and fixes build issues on Rust 1.51.

Fixes #36

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -13,6 +13,7 @@ use matrix_sdk::{
                 RedactedMessageEventContent, ServerNoticeMessageEventContent,
                 TextMessageEventContent, VideoMessageEventContent,
             },
+            EncryptedFile,
         },
         RedactedSyncMessageEvent, SyncStateEvent,
     },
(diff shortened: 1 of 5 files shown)


src/render.rs
```diff
@@ -310,18 +311,106 @@ impl Render for ServerNoticeMessageEventContent {
...
+    if url.scheme() != "mxc" {
+        return Err("URL missing MXC scheme".into());
+    }
+
+    if url.path().is_empty() {
+        return Err("URL missing path".into());
+    }
+
+    Ok(format!(
+        "{homeserver}_matrix/media/r0/download/{server_name}{media_id}",
```
poljar, line 330: I think using [join](https://docs.rs/url/2.2.1/url/struct.Url.html#method.join) for this is a better idea.

src/render.rs
```diff
@@ -310,18 +311,106 @@ impl Render for ServerNoticeMessageEventContent {
...
+    let host_str = format!(
+        "emxc://{}",
+        homeserver
+            .host_str()
+            .ok_or("Missing homeserver host string")?
+    );
+    let emxc_homeserver = url::Url::parse(&host_str)?;
+    let server_name = url.host().ok_or("Missing host")?;
+
+    let base_url = format!(
```
poljar, line 372: Same as above, using join makes this a bit easier to follow and adds a bit more type safety.

poljar (changes requested):
Thanks for working on this, a couple of changes are needed before we merge this.

Austin-Ray:
I've replaced with `format!()` with `Url.join()`; however, chaining them kept overwriting part of the path so I moved some into the join and kept a smaller format. Am I missing an obvious cause for the overwrite?

I'm looking into replacing that smaller `format!()` with a function call since both `mxc_to_http()` and `mxc_to_emxc()` use the same form. 

Also, add two small tests for faster dev since the plugin takes a while to connect.

poljar:
> I've replaced with `format!()` with `Url.join()`; however, chaining them kept overwriting part of the path so I moved some into the join and kept a smaller format. Am I missing an obvious cause for the overwrite?

The docs for `join()` mention this:

> Note: a trailing slash is significant. Without it, the last path component is considered to be a “file” name to be removed to get at the “directory” that is used as the base

In other words this will overwrite:

```rust
    let url = base.join("test")?.join("another")?;
```

While this won't:

```rust
    let url = base.join("test/")?.join("another/")?;
```
> I'm looking into replacing that smaller `format!()` with a function call since both `mxc_to_http()` and `mxc_to_emxc()` use the same form. However, that might be pre-mature abstraction that'll make it harder to understand.

I think that's a good idea, the methods are simple enought so I don't think it's going to be hard to understand what's going on.

> Also, add two small tests for faster dev since the plugin takes a while to connect.

Tests are always welcome, especially when it's easy to write them, i.e. when we have pure functions like this one.

Austin-Ray:
> I think that's a good idea, the methods are simple enought so I don't think it's going to be hard to understand what's going on.

Refactored into dedicated method.

On the overwriting part, it turns out it's from the `url.path()` call. It returns `/media-id` and when joining it a leading `/`, it replaces the entire path.  Since the Matrix docs specify an `mxc` URI to only have `media-id` in its path, should I modify the code to extract `media-id` with `url.path_segments()` or is the `format!()` sufficiently safe?

[0] https://matrix.org/docs/spec/client_server/latest#matrix-content-mxc-uris

poljar:
> On the overwriting part, it turns out it's from the `url.path()` call. It returns `/media-id` and when joining it a leading `/`, it replaces the entire path. Since the Matrix docs specify an `mxc` URI to only have `media-id` in its path[0], should I modify the code to extract `media-id` with `url.path_segments()` or is the `format!()` sufficiently safe?

Ah ok, this will become easier once we update to the latest rust-sdk/Ruma, since Ruma got a specialized type for MXC URLs: https://github.com/ruma/ruma/blob/main/ruma-identifiers/src/mxc_uri.rs#L17.

Let's leave it with the format for now.

Austin-Ray:
> Let's leave it with the format for now.

Sounds good! Any other changes or am I good to rebase the history so it's tidy?

poljar:
No further changes needed, thanks.

Austin-Ray:
Tidied the history and CI is good

PR #39, opened 2021-04-08
doc: remove nightly rust requirement

Austin-Ray:
Commit 43e6f20 allowed building with stable Rust so remove the
requirement from `README.md`.

Just something small I noticed while looking over the repo.


poljar:
Thanks.

PR #56, opened 2021-07-29
Verification suport

### LINKED ISSUE #32: Support emoji verification.
While the script supports only to-device emoji verification we should probably support in-room verification as well.

poljar:
This PR adds support for interactive verification, this needs a couple of things for everything to work correctly:

- [x] React to verification requests over to-device messaging
- [x] React to verification requests in DMs
- [ ] Command to request self verifications (this is over to-device messaging)
- [ ] Command to request verification of other users
- [x] Command to accept/cancel/confirm active verifications
- [x] Support rendering of emoji verifications
- [x] Support rendering of QR code verifications

weechat-emoji
weechat-qr

The command to control an active verification is called `/verification` it has the `accept`, `confirm`, `cancel`, and `use-emoji` subcommands.

The plan for the command that enables verifications to be requested should be called `/verify`:

`/verify <user_id> [<device_id>] [-fingerprint <device fingerprint>]`.

This would allow us to request verification with a user, or with a specific device if the device id is given, alternatively if we provide a fingerprint using the `-fingerprint` switch a user or device is manually marked as verified.

This closes: #32.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "addr2line"
-version = "0.14.1"
+version = "0.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a55f82cfe485775d02112886f4169bde0c5894d75e79ead7eafe7e40a25e45f7"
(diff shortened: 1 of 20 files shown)


panekj:
Fixed via #89

PR #58, opened 2021-08-02
Added type localvar and updated github action to include ubuntu-latest

pvonmoradi:
localvar `type` is introduced. For DMs, it is set to `private` and for non-DMs, it is set to `channel`.  
Note that I'm not familiar with Rust; I've added `is_direct` to `MatrixRoom` and I assume it uses https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_base/struct.Room.html#method.is_direct 

Also, the workflow is updated to upload the binary directly to the Releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,11 +8,14 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04]
+        os: [ubuntu-18.04, ubuntu-latest]
         include:
           - os: ubuntu-18.04
             artifact_name: libmatrix.so
             asset_name: libmatrix.so-ubuntu-18.04
+          - os: ubuntu-latest
+            artifact_name: libmatrix.so
+            asset_name: libmatrix.so-ubuntu-latest
         rust:
           - stable
       fail-fast: false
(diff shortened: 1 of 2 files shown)


catthehacker:
Releases are not the same thing as artefacts, please don't change the way it's done now
Releases require a tag on the repo, and they will show up via releases API. Artefacts are still available to download but are not a public release and they are not changing repo.

pvonmoradi:
@catthehacker I think the my `github:` commit could be ignored if we don't want to use Releases API. 
About title, misleading how? Doing two things in one PR?

catthehacker:
> About title, misleading how? Doing two things in one PR?

imo, I (try to) treat PR names as squashed commits brief description. In case of this PR, because of notification I expected only Rust source to change (2nd commit only).

catthehacker:
> I think the my `github:` commit could be ignored if we don't want to use Releases API.

As for this, I don't have a problem with Releases API, but every PR will become a release. I don't think that's a reasonable thing to do 😺 
Best case scenario, we would like to have artefacts produced by `push` and `pull_request` events, and releases produced by tagged `push` events.

poljar:
> > I think the my `github:` commit could be ignored if we don't want to use Releases API.
> 
> As for this, I don't have a problem with Releases API, but every PR will become a release. I don't think that's a reasonable thing to do smiley_cat
> Best case scenario, we would like to have artefacts produced by `push` and `pull_request` events, and releases produced by tagged `push` events.
> _but obviously this is my opinion as a dev/user, if you/projects feels like it, go for it_

Yeah we'll want to have releases just for pushed tags, though nightly releases are fine for pushes to master.

pvonmoradi:
> > > I think the my `github:` commit could be ignored if we don't want to use Releases API.
> > 
> > 
> > As for this, I don't have a problem with Releases API, but every PR will become a release. I don't think that's a reasonable thing to do smiley_cat
> > Best case scenario, we would like to have artefacts produced by `push` and `pull_request` events, and releases produced by tagged `push` events.
> > _but obviously this is my opinion as a dev/user, if you/projects feels like it, go for it_
> 
> Yeah we'll want to have releases just for pushed tags, though nightly releases are fine for pushes to master.

I removed the rule that pushes to Releases and added the former artifact upload action back. Also added ubuntu-latest to build list.

poljar (approved):
So this is fine but it doesn't update the `type` localvar when we receive a new [`m.direct` event](https://docs.ruma.io/ruma/events/enum.AnyGlobalAccountDataEvent.html#variant.Direct).

Since the type of the room doesn't change often and a restart would fix the types again we can leave the dynamic updating for later.

PR #69, opened 2022-04-17
Makefile support for separate debug and release build targets

dkasak:


MrIceman (approved):

strk:
I think GNU Coding Standards call for an `install-strip` to install the non-debug version ?

lnicola:
I think the binary targets should be phony too. The current `SOURCES` is broken because it's missing two submodules (and also the manifest), and `cargo build` already knows what to do.

lnicola:
> I think GNU Coding Standards call for an install-strip to install the non-debug version ?

That doesn't work directly because:

 - by default, Rust keeps debug info in debug builds and strips it in release ones
 - this can be overridden in the manifest (e.g. `[profile.release] strip = "none"` in `Cargo.toml`) or in other like environment variables
 - thus, whether an artifact has debug info is a property of how the artifact was built, not of how it is installed

Of course, you can enable debug info in `[profile.release]` and then strip it in `make install-strip`. But it will slow down the builds in this case.

strk:
Isn't `install -s` as effective ?

https://man7.org/linux/man-pages/man1/install.1.html

lnicola:
Just as effective as `strip`, yes, but a bit slower than letting Cargo do it.

PR #72, opened 2022-10-02
add socks feature

### LINKED ISSUE #47: Support for proxy (SOCKS5)
Wondering if it's possible to add a configuration option for a proxy (or even multiple proxies), specifically SOCKS5, but I would assume that adding support for others might not be a problem if they're prefixed correctly in the config

badlandspray:
Fixes #47


poljar (approved):
Thanks, looks good.

PR #88, opened 2024-10-14
Fix GitHub CI & add ubuntu-22.04

panekj:


panekj:
Hm, I was confused why I did get build failure[1] but it's not showing up in this repo only to realise that `ubuntu-24.04` is just now becoming `ubuntu-latest`[2] (means `ubuntu-latest` resolves currently to both `ubuntu-24.04` and `ubuntu-22.04` depending on repository or CI availability).

https://github.com/actions/runner-images/issues/10636

[1]: https://github.com/panekj/weechat-matrix-rs/actions/runs/11329358081/job/31504708785
[2]: https://github.com/actions/runner-images/#available-images

poljar (approved):
Oh my, I would have expected them to bump things atomically. I guess they roll things out gradually in case something goes wrong.

panekj:
They always used rolling strategy, I just have not been keeping up with updates to GH Actions so I was unprepared and questioning my sanity (also panicking for 5 minutes where the failure occurred because I did not notice it was in my fork)

PR #91, opened 2025-02-26
Use XDG_CONFIG_HOME for the default WEECHAT_HOME

strk:
Modern systems use that new path. For sure I know weechat 4.5.1 does, not sure since when

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-WEECHAT_HOME ?= $(HOME)/.weechat
+XDG_CONFIG_HOME ?= $(HOME)/.config
+WEECHAT_HOME ?= $(XDG_CONFIG_HOME)/weechat
 PREFIX ?= $(WEECHAT_HOME)
 
 SOURCES := $(wildcard src/*.rs src/commands/*.rs Cargo.lock)


poljar (approved):
Thanks.

Makefile
```diff
@@ -1,4 +1,5 @@
-WEECHAT_HOME ?= $(HOME)/.weechat
+XDG_CONFIG_HOME ?= $(HOME)/.config
+WEECHAT_HOME ?= $(XDG_CONFIG_HOME)/weechat
```
trygveaa, line 2: This is not correct for plugins. It will install it into `~/.config/weechat/plugins`, but when XDG directories are used it should be installed in `~/.local/share/weechat/plugins`.

Additionally, people with an existing config that haven't explicitly switched to XDG directories will still be using `~/.weechat`. WeeChat will use `~/.weechat` if that exists and `~/.config/weechat/` does not, and it will use XDG directories if `~/.weechat` doesn't exist or `~/.config/weechat/` exists.

strk: What happens if both `~/.weechat` and `~/.config/weechat` exist ?
Why is `~/.config/weechat` checked at all if the plugin install directory is `~/.local/share/weechat/plugins` ?

Is there an env variable that weechat uses similar to `WEECHAT_HOME` but for plugins ? Or is the presence of a `WEECHAT_HOME` a sign that `$(WEECHAT_HOME)/plugins` is the place we want plugins installed ?

trygveaa: > What happens if both ~/.weechat and ~/.config/weechat exist ?

Then `~/.config/weechat` is used.

> Why is ~/.config/weechat checked at all if the plugin install directory is ~/.local/share/weechat/plugins ?

Because when using XDG directories it's spread over multiple directories. Config is in `~/.config/weechat`, but data like plugins is in `~/.local/share/weechat`. WeeChat only checks the config directory to determine if `~/.weechat` or XDG directories should be used.

> Is there an env variable that weechat uses similar to WEECHAT_HOME but for plugins ? Or is the presence of a WEECHAT_HOME a sign that $(WEECHAT_HOME)/plugins is the place we want plugins installed ?

Right, so what I said only applies if `WEECHAT_HOME` is not set. If `WEECHAT_HOME` is set, then `$(WEECHAT_HOME)/plugins` is correct to use. There is no other WeeChat specific env variables that controls this as far as I know, but WeeChat will adhere to `XDG_CONFIG_HOME` and `XDG_DATA_HOME` when XDG directories are used.

trygveaa: > There is no other WeeChat specific env variables that controls this as far as I know

Actually, I found that there's one more. If `WEECHAT_EXTRA_LIBDIR` is set, weechat will also load plugins from `${WEECHAT_EXTRA_LIBDIR}/plugins`. But this is in addition to the directories it normally loads plugins from, so probably not relevant to this makefile.

PR #94, opened 2025-03-18
weechat: fix listing devices

panekj:
Smaller PR broken out of feat/verification to get it faster to into main


poljar (approved):

JRGonz:
Ah ok. So this was for device listing and not cross signing verification?

Issue #78, opened 2022-12-14
No more message received after a deserialization error


erdnaxeli:
I just got this error
```
2022-12-14T16:33:35.645114Z ERROR matrix_sdk::sync: Received an invalid response: deserialization failed: ID exceeds 255 bytes at line 1 column 176847
```

After that no communication (both ways) was working anymore.

edit: the messages are actually correctly sent, but I got no feedback. Only the reception is broken.

PR #103, opened 2025-05-04
connection: run sync in loop, break on Ok

panekj:
Client sync exits with Err() in case of any sync issues
and Ok() if it's gracefully shutdown, this fixes a problem
when weechat-matrix-rs stops receiving messages
(most likely due to network connection issue)

Fixes https://github.com/poljar/weechat-matrix-rs/issues/78

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -428,8 +428,9 @@ impl Connection {
 
         let client_ref = &client;
 
-        let _ret = client
-            .sync_with_callback(sync_settings, |response| async move {
+        loop {
+            let ret = client
+            .sync_with_callback(sync_settings.clone(), |response| async move {
                 for (room_id, room) in response.rooms.join {
                     for event in
                         room.state.iter().filter_map(|e| e.deserialize().ok())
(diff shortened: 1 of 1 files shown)


src/connection.rs
```diff
@@ -428,8 +428,9 @@ impl Connection {
 
         let client_ref = &client;
 
-        let _ret = client
-            .sync_with_callback(sync_settings, |response| async move {
+        loop {
+            let ret = client
+            .sync_with_callback(sync_settings.clone(), |response| async move {
```
poljar, line 433: There's nowadays a `SyncServic` which handles this looping automatically.

I think it'll make sense to eventually migrate over to it: https://docs.rs/matrix-sdk-ui/latest/matrix_sdk_ui/sync_service/struct.SyncService.html#example.

poljar (approved):

src/connection.rs
panekj: I've seen the matrix_sdk_ui crate but I haven't explored it (yet, also it's been so long that I was used to just matrix_sdk). I'll try to experiment with it at some point.

poljar: It builds on top of the `matrix-sdk` crate and provides some higher level and mostly UI client specific functionality.

PR #104, opened 2025-05-04
commands: add /me

panekj:


src/commands/me.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn callback(
+        &mut self,
+        _: &Weechat,
+        buffer: &Buffer,
+        cmd: Cow<str>,
+    ) -> ReturnCode {
+        if let Some(room) = self.servers.find_room(buffer) {
+            self.servers.runtime().block_on(room.send_message(
+                RoomMessageEventContent::emote_plain(
+                    cmd.strip_prefix("/me ").map(|s| s.to_string()).unwrap(),
```
panekj, line 40: Could also use split args and drop first element, no idea which is better

poljar: I marginally like the `strip_prefix` more, then the rest could be parsed by `clap`.

Obviously that's not needed here.

poljar (approved):

PR #106, opened 2025-05-09
weechat: upgrade clap, add multiple commands

panekj:


strk:
Anything useful left in this PR ? The multiple commands I believe were already implemented in other PRs

PR #114, opened 2025-10-09
chore: unify tracing usage

### LINKED ISSUE #48: The Matrix Space idea - Recognize which rooms are under a given Space
Correct me if I am wrong but currently the weechat ties channels to servers. It would be nice to tie a room/channel to space(s) it is part of using the same hierarchical scheme the server and the channel use. Why? Because then one could use a plugin such as the collapse_channel.py https://weechat.org/scripts/source/collapse_channel.py.html/ to show only part of the channels on the buflist and keep it "tidy" that way.  It would add convenience as you could group big number of rooms to any number of groups one wish for by using private Spaces.

If Matrix is to grow towards the direction of Discord, there will be Spaces with rooms for each subtopic in a Space, and eventually the number of rooms one is connected grows large. Being able to expand/collapse them in a more manual manner is not possible right now and this "simple" trick would solve these problems.

panekj:


strk:
Are you still interested in getting this PR merged ? If so can you please describe what the goal is ?

PR #119, opened 2026-01-20
Add SSO login support

bogdando:
Changes:
1. Updated Cargo.toml Added "sso-login" feature to the matrix-sdk dependency

2. Updated connection.rs
Added a local HTTP server to automatically capture the SSO callback.
Added SsoLoginUrl(String) variant to ClientMessage enum.
Added LoginType import from ruma::api::client::session::get_login_types::v3.
Updated response_receiver to handle SsoLoginUrl messages Modified sync_loop to:
  Query available login flows using get_login_types() Check if SSO is available
  Generate and display SSO URL when SSO is available and credentials are empty
  Fall back to password login when credentials are provided.
Added login_with_token() method to complete SSO login with a token.

4. Updated server.rs 
Added receive_sso_url() method to display SSO URL and optionally open browser.
Added complete_sso_login() method to complete SSO login with a token.
Added use_sso parameter (defaults to False).

6. Updated commands/matrix.rs
Added sso-complete command to the command structure.
Added handler method sso_complete_command().
Added subcommand to argparse.

How to use:
* Connect with empty username/password (and use_sso on): /matrix connect myserver

The plugin will:
Check if SSO is available on matrix.org
Display the SSO login URL
Optionally open it in your browser

* After "completing" SSO in the browser (the connection failure is expected): Copy the loginToken from the callback URL (the parameter after loginToken=)

* Run: ``/matrix sso-complete m <login-token>``
The plugin will complete the login and you'll be connected.

The implementation follows the same pattern as the matrix-rust-sdk examples and should work with matrix.org and other homeservers that support SSO.

Note that the sessions list in the WEB client will be showing the connected device as unverified, which still works for connections not leveraging E2E encryption (https transport encrpytion would still apply).

Assisted-by: Cursor AI agent (Claude 4.5 Opus)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -242,6 +242,58 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
+name = "axum"
+version = "0.8.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b52af3cb4058c895d37317bb27508dccc8e5f2d39454016b297bf4a400597b8"
+dependencies = [
+ "axum-core",
+ "bytes",
+ "form_urlencoded",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
+ "itoa",
+ "matchit",
+ "memchr",
+ "mime",
+ "percent-encoding",
+ "pin-project-lite",
+ "serde_core",
+ "serde_json",
+ "serde_path_to_error",
+ "serde_urlencoded",
+ "sync_wrapper",
+ "tokio",
+ "tower",
+ "tower-layer",
+ "tower-service",
+ "tracing",
+]
+
+[[package]]
+name = "axum-core"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08c78f31d7b1291f7ee735c1c6780ccde7785daae9a9206026862dab7d8792d1"
+dependencies = [
+ "bytes",
+ "futures-core",
+ "http",
+ "http-body",
+ "http-body-util",
+ "mime",
+ "pin-project-lite",
+ "sync_wrapper",
+ "tower-layer",
+ "tower-service",
+ "tracing",
+]
+
+[[package]]
 name = "backon"
 version = "1.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


bogdando:
It works for me with weechat v4.8.1, and matrix.org account.
I did not check if this change does not regress non sso login cases (use_sso off, and non empty userpass or username) as I don't have a server to try that, sorry

andyrtr:
I want to give it a try with my matrix.org and archlinux.org sso accounts. I'm getting the localhost [not connected] and the SSO session will timeout. How do I setup the local http server?

bogdando:
> I want to give it a try with my matrix.org and archlinux.org sso accounts. I'm getting the localhost [not connected] and the SSO session will timeout. How do I setup the local http server?

the local http server is started by the plugin code. The failed connection is expected, you should only copy the loginToken from the address URL, and use it with sso-complete, as documented in the commit message. Thanks for trying, let me know how it goes for you

andyrtr:
Thanks. Login works for Arch SSO. 

Problem here: all connected channels are listed at the end is seems and not connected to the matrix/Server account and are wildly mixed with other irc channels.

bogdando:
> Thanks. Login works for Arch SSO.
> 
> Problem here: all connected channels are listed at the end is seems and not connected to the matrix/Server account and are wildly mixed with other irc channels.

ouch, that's unfortunate. I am not certain I could help to address that, just the 1st time trying this thing :)
there is another [PR](https://github.com/poljar/weechat-matrix-rs/pull/120) to implement ```/matrix join```, maybe it works better when joining from the weechat instead of WEB client? IIRC, weechat also allows manually sorting/reordering of the list of the channels

UPDATE: I tried
```
/script install autosort.py
/set irc.look.server_buffer independent
```
and it looks much better already, @andyrtr

andyrtr:
After each /quit and restart and /matrix connect ServerName the SSO has to be done again. Is there any way to permanently store it? A /save doesn't do this for me.

bogdando:
> After each /quit and restart and /matrix connect ServerName the SSO has to be done again. Is there any way to permanently store it? A /save doesn't do this for me.

@andyrtr it could be, but implementing that would become another degree of complexity to me. We could start with that basic support and improve.

andyrtr:
I can't get autosort script to work here with the matrix-rs plugin Accounts. Autosort works with two znc-libera/znc-oftc buffers but password login for Matrix.org account and my sso ArchLinux account won'r get listed in a separate buffer. But that's unrelated to sso.

Have you had a look at [iamb sso](https://github.com/ulyssa/iamb/pull/160) ?

Komzpa:
I opened #154 as a smaller SSO replacement.

It uses the Matrix SDK `sso-login` flow and its callback handling, instead of adding a plugin-local manual `/matrix sso-complete` token path. Manual token entry still sounds useful, but I would keep it as a follow-up after the SDK SSO path lands.

Fix requested by @strk.

bogdando:
closing in the preference of https://github.com/poljar/weechat-matrix-rs/pull/154

PR #120, opened 2026-01-20
Add matrix join command for a server

bogdando:
Add ``/matrix join <server> <channel>``

Assisted-by: Cursor AI agent (Claude 4.5 Opus)


strk:
Why would the server name be needed ?

Komzpa:
I agree the server argument should not be needed in normal room-buffer use.

#131 adapts this PR without the extra `<server>` argument: `/matrix join <room-id-or-alias>` works, and direct `/join <room-id-or-alias>` works from Matrix buffers.

Fix requested by @strk.

bogdando:
closing for the preference of https://github.com/poljar/weechat-matrix-rs/pull/131

Issue #124, opened 2026-06-10
Fails to load against weechat 4.9.1


kugel-:
`API mismatch for plugin "/home/u1000/.local/share/weechat/plugins/matrix.so" (current API: "20260530-01", plugin API: "20240727-01"), failed to load`

I suspect that `weechat-sys` is stale?

kugel-:
I checked out rust-weechat and pointed Cargo.toml to that. 

Although it says "using system header", it effectively uses the vendored header.

``` 
 main = origin/main$ cargo build
   Compiling weechat-sys v0.4.0 (/home/u1000/dev/rust-weechat.git/crates/weechat-sys)
warning: weechat-sys@0.4.0: Using system header via wrapper.h
   Compiling weechat v0.4.0 (/home/u1000/dev/rust-weechat.git/crates/weechat)
warning: weechat@0.4.0: WEECHAT_PLUGIN_API_VERSION: 20240727
   Compiling weechat-matrix v0.1.0 (/home/u1000/dev/weechat-matrix-rs.git)
warning: unused variable: `ret`
``` 

I must add that `weechat-plugin.h` is not in a standard location since I installed it via `nix-env`. But the above happens even with `PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig/ cargo build`.

Only if I actually replace the bundled weechat-plugin.h the plugin can successfully load (except that that I need to use `patchelf` to add `/usr/lib/x86_64-linux-gnu` to the library search path because otherwise libsqlite3.so.0 can't be found).

strk:
I'm also having this problem. Maybe we should DROP the bundled -dev and just fail unless the builder explicitly set the header file to use. What do you think @kugel- ? @poljar ?

@Komzpa your opinion ?

Komzpa:
I agree with making the bundled header explicit rather than a silent fallback. The behavior lives in `weechat-sys`, so I opened poljar/rust-weechat#48 for that: default to the system header or a clear failure, and keep `WEECHAT_BUNDLED=true` as the explicit bundled-header opt-in.

strk:
@kugel- I merged #48 - you ok with the outcome of this ? Does it make your `API mismatch` issue go away ? If so please close this, or give feedback.

kugel-:
@strk sorry for the late reply. Yes, that works thanks.

[...]

PR #123, opened 2026-05-21
fix: build on rustc 1.94 and newer

toothrot:
The default compile time recursion_limit was lowered, and matrix-sdk depends on higher limits.

Fixes poljar/weechat-matrix-rs#122

Issue #122, opened 2026-02-20
Build Error E0275


JRGonz:
```
   Compiling weechat-matrix v0.1.0 (/src/weechat-matrix-rs)                                                                                                    
error[E0275]: overflow evaluating the requirement `(std::string::String, matrix_sdk::encryption::vodozemac::olm::DecryptionError): std::marker::Send`                                 
    --> src/connection.rs:128:17                                                                                                                                                      
     |                                                                                                                                                                                
 128 |         runtime.spawn(Connection::sync_loop(                                                                                                                                   
     |                 ^^^^^                                                                                                                                                          
     |                                                                                                                                                                                
     = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`matrix`)                   
```

flips:
I started seeing the same after `rustup` a good while back …
But reading that further up, I'm not sure if that's the reason build fails, as it now ends with:

```
    --> /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.47.1/src/runtime/runtime.rs:232:21
     |
 230 |     pub fn spawn<F>(&self, future: F) -> JoinHandle<F::Output>
     |            ----- required by a bound in this associated function
 231 |     where
 232 |         F: Future + Send + 'static,
     |                     ^^^^ required by this bound in `Runtime::spawn`
     = note: the full name for the type has been written to '/home/myusername/src/git/weechat-matrix-rs/target/release/deps/matrix.long-type-2636114478810706611.txt'
     = note: consider using `--verbose` to print the full type name to the console
     = note: this error originates in the attribute macro `instrument` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0275`.
error: could not compile `weechat-matrix` (lib) due to 1 previous error
```
Maybe I should open a separate issue? :)

toothrot:
There's a bit of related discussion on the matrix sdk: https://github.com/matrix-org/matrix-rust-sdk/pull/3749#issuecomment-2312939823

I've played with this a bit locally, trying to update us to a newer SDK version, but we'll still run into this issue for now. 

Since this isn't a library that will be depended on by other rust projects, it may not be a bad idea to add the suggested `#![recursion_limit = "256"]` in `lib.rs`. I'll send a pull request after testing it.

[...]

PR #118, opened 2026-01-20
Fix plugin to work with weechat 4.1.0+

bogdando:
Make sure it works with v4.8.1.

PR #127, opened 2026-06-23
fix: build with current WeeChat APIs and rustc 1.94+

Komzpa:
This finishes the current-WeeChat API build fix on top of merged #123 and merged rust-weechat#46/#47.

It folds in #118 with Bogdan Dobrelya's authorship preserved, adds a small `build.rs` so this crate gets the right WeeChat API cfg, and points back at upstream `poljar/rust-weechat` now that the needed rust-weechat fixes have landed.

Related to #90, #117, and #124. Compile checks pass against both bundled and installed WeeChat headers, and a headless WeeChat 4.6.3 smoke loaded the built `matrix.so` successfully. #122 is already fixed upstream by #123.

Checked after updating to upstream rust-weechat commit `adb96ffc9f229cbff05ad51b0ff4c9d46597e2f4`:

```text
cargo fmt --check
git diff --check
docker run --rm -v "$PWD":/work -w /work rust:1.96-bookworm sh -lc 'apt-get update >/dev/null && apt-get install -y --no-install-recommends libclang-dev >/dev/null && WEECHAT_BUNDLED=1 CARGO_TARGET_DIR=/tmp/weechat-127-post47-bundled /usr/local/cargo/bin/cargo check --locked'
docker run --rm -v "$PWD":/work -w /work rust:1.96-bookworm sh -lc 'apt-get update >/dev/null && apt-get install -y --no-install-recommends libclang-dev weechat-dev >/dev/null && WEECHAT_PLUGIN_FILE=/usr/include/weechat/weechat-plugin.h CARGO_TARGET_DIR=/tmp/weechat-127-post47-system /usr/local/cargo/bin/cargo check --locked'
```

Current head `aea50f27f0fb84a372c9883a30d58fffaebe90ac` passed GitHub Actions on ubuntu-22.04, ubuntu-24.04, and ubuntu-latest.

Fix requested by @strk.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,23 +11,23 @@ dependencies = [
  "macroific",
  "proc-macro2",
  "quote",
- "syn 2.0.87",
+ "syn 2.0.114",
 ]
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aead"
@@ -45,16 +45,16 @@ version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
- "cfg-if 1.0.3",
+ "cfg-if 1.0.4",
  "cipher",
  "cpufeatures",
 ]
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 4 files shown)


Komzpa:
@strk Could you please review this draft when you have a moment? I prepared it by your request.

strk-ai-agent (approved):
## Overview

This PR fixes builds against both old (4.1.x) and current (4.2.x+) WeeChat API headers, and also resolves compilation on rustc 1.94+. The approach is sound: a `build.rs` script probes the WeeChat API version at compile time and sets cfg flags, then `config.rs` uses those flags to call the correct API.

CI passes on ubuntu-22.04, ubuntu-24.04, and ubuntu-latest. The smoke test (loading `matrix.so` into headless WeeChat 4.6.3) succeeded.

---

## Detailed Review

### `Cargo.toml` ✓
- Adding `weechat-sys` as a build dependency is necessary and correct — `build.rs` needs access to `WEECHAT_PLUGIN_API_VERSION` at compile time.
- No version constraint is needed since the resolved revision comes from the lockfile (same `poljar/rust-weechat` repository).

### `build.rs` ✓
- Properly uses `cargo::rerun-if-changed` and `cargo::rerun-if-env-changed` for incremental correctness.
- Registers cfg flags with `cargo::rustc-check-cfg` — good practice for modern Rust.
- Parses the version string and sets the appropriate cfg flag. The logic is straightforward and correct.
- Minor nit: the `#[allow(unused)]` on the `WeechatApiVersions` enum is unnecessary — the enum variants **are** used in the comparison (`WeechatApiVersions::V4_2_0 as u64`). But this is cosmetic and harmless.

### `src/config.rs` ✓
- Conditional imports with `#[cfg(weechat410)]` / `#[cfg(not(weechat410))]` correctly select the right option type.
- The `LookSection::server_buffer()` refactor from `if let` to a `match` with cfg-guarded arms is cleaner and more maintainable (the `_ => panic!` catch-all is good defense).
- **Design note:** The two option-creation blocks are near-duplicates (differing only in `IntegerOptionSettings`/`new_integer_option` vs `EnumOptionSettings`/`new_enum_option`). This is acceptable for a build-fix PR — keeping it explicit avoids over-engineering. Could be deduplicated later with a helper macro if desired.

### `Cargo.lock` ✓
- Pins `weechat`/`weechat-macro`/`weechat-sys` to the upstream revision (`adb96ffc`) that contains the fixed `rust-weechat` API.
- All other bumps are transitive dependency updates from the SDK; no concerns.

---

## Suggestions (minor, non-blocking)

1. **`build.rs`:** Remove `#[allow(unused)]` — the enum variants are used. Or alternatively, use a named constant instead of an enum if you want to suppress the lint differently.
2. **`config.rs`:** The duplicated config-option block could be abstracted (e.g., a local macro or a generic helper), but this is purely optional and I would consider it future cleanup, not a PR blocker.

---

## Verdict

**Approved.** The changes are correct, minimal, well-tested, and solve the stated problem. All three CI jobs pass. Thank you for including both the `weechat410` (old API) and `weechat420` (new API) paths and for the careful smoke testing.

PR #139, opened 2026-06-23
fix: add smart filter tags for membership events

### LINKED ISSUE #22: Smart filter support.
The Python script somewhat supports smart filtering like the IRC plugin does, it is a bit flaky but often times it works well enough to hide massive amounts of joins/leaves.

To support this we should track the activity of members, e.g. every time we fetch a member to render an event that is send by that member we should save a timestamp of their last activity. Then based on that timestamp add a smart filter tag when we render membership events.

Komzpa:
Adds smart-filter tags to membership events when the affected member has not been active recently, and documents the filter command.

Activity is now updated from message-like and state events, not just rendered message bodies, and the inactivity delay is configurable with `matrix-rust.look.smart_filter_delay` (default five minutes).

The final filter decision still uses the affected member, matching #22's join/leave-noise case; sender activity is tracked but does not decide whether another member's kick/ban line is hidden.

Fixes #22.

Fix requested by @strk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,11 @@ _(replace the placeholders in brackets [] with your own details)_:
 
        /save
 
+To hide join/leave noise from users who have not spoken recently, add a WeeChat
+filter for Matrix smart-filter tags:
+
+       /filter add matrix_smart * matrix_smart_filter *
+
 
 # Helpful Commands
 
--- a/README.md
+++ b/README.md
@@ -117,10 +117,14 @@ _(replace the placeholders in brackets [] with your own details)_:
 
        /save
 
-To hide join/leave noise from users who have not spoken recently, add a WeeChat
+To hide join/leave noise from users who have not been active recently, add a WeeChat
 filter for Matrix smart-filter tags:
 
        /filter add matrix_smart * matrix_smart_filter *
+
+The inactivity delay defaults to five minutes and can be changed with:
+
+       /set matrix-rust.look.smart_filter_delay 300000
 
 
 # Helpful Commands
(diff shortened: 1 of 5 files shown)


strk-ai-agent:
I reviewed the PR — overall looks clean and correct, thank you.

I have a few suggestions for improvement:

1. **Trigger `mark_active` on more event types** — Currently only message content (text/emote/notice/etc.) updates the "last active" timestamp. It would also make sense to trigger it on reactions, edits, redactions, and state events, since those are also signs of user activity.

2. **Check `sender_id` instead of `target_id` in `should_smart_filter`** — For events like kicks and bans, the target is the person being acted upon, but the *sender* is the one actively participating. If a moderator who has spoken recently kicks a user who has never spoken, the kick event gets filtered (because the target has no activity). It seems more useful to check the sender, since the sender is the one taking action in the room.

3. **Make the inactivity threshold configurable** — The 300000 ms (5 min) constant is reasonable but hardcoded. A config option (e.g. `matrix-rust.look.smart_filter_threshold`) would let users tune it to their preference.

Would you prefer to merge the PR as-is and address these in follow-ups, or are you interested in working on these improvements before merging?

Either way is fine — just let me know what you think.

Komzpa:
Picked up the activity-source and configurable-delay parts in the updated branch.

I kept the final filter decision on the affected member rather than the event sender: #22 is about hiding join/leave-style membership noise for inactive members, and a moderator action still renders a line about the target member. The sender still updates activity now, so direct sender membership events and later activity are tracked.

strk:
Lovely, thank you so much !

PR #145, opened 2026-06-23
feat: filter exported keys by room

### LINKED ISSUE #33: Allow export of subsets of room encryption keys based on a predicate
The most important feature would be to export room encryption keys per room. This would allow, for instance, an old member of an E2E encrypted room to supply keys to a newer member so he can read history prior to his join.

The function [export_keys](https://github.com/matrix-org/matrix-rust-sdk/blob/4f2cad8f626068f05d567588d8398d0952536a46/matrix_sdk_crypto/src/machine.rs#L1122) from the Rust SDK could be used for the implementation.

Komzpa:
Adds `/keys export --room-id ROOM_ID <file> <passphrase>` so an E2EE key export can be limited to one room instead of always exporting every room key.

Fixes #33.

Fix requested by @strk.

### DIFF
--- a/src/commands/keys.rs
+++ b/src/commands/keys.rs
@@ -4,6 +4,7 @@ use clap::{
     App as Argparse, AppSettings as ArgParseSettings, Arg, ArgMatches,
     SubCommand,
 };
+use matrix_sdk::ruma::{OwnedRoomId, RoomId};
 
 use weechat::{
     buffer::Buffer,
@@ -32,7 +33,7 @@ impl KeysCommand {
         let settings = CommandSettings::new("keys")
             .description(Self::DESCRIPTION)
             .add_argument("import <file> <passphrase>")
-            .add_argument("export <file> <passphrase>")
+            .add_argument("export [--room-id <room-id>] <file> <passphrase>")
             .arguments_description(
                 "file: Path to a file that is or will contain the E2EE keys export",
             )
@@ -72,9 +73,14 @@ impl KeysCommand {
         Weechat::spawn(import()).detach();
     }
 
-    fn export(server: MatrixServer, file: PathBuf, passphrase: String) {
+    fn export(
+        server: MatrixServer,
+        file: PathBuf,
+        passphrase: String,
+        room_id: Option<OwnedRoomId>,
+    ) {
         let export = || async move {
-            server.export_keys(file, passphrase).await;
+            server.export_keys(file, passphrase, room_id).await;
         };
         Weechat::spawn(export()).detach();
     }
@@ -89,10 +95,13 @@ impl KeysCommand {
                     Self::import(server, file, passphrase);
                 }
                 ("export", args) => {
-                    let (file, passphrase) = Self::upcast_args(
-                        args.expect("No args were provided to the subcommand"),
-                    );
-                    Self::export(server, file, passphrase);
+                    let args =
+                        args.expect("No args were provided to the subcommand");
+                    let (file, passphrase) = Self::upcast_args(args);
+                    let room_id = args
+                        .value_of("room-id")
+                        .map(|room_id| RoomId::parse(room_id).unwrap());
+                    Self::export(server, file, passphrase, room_id);
                 }
                 _ => unreachable!(),
             }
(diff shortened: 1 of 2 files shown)


Komzpa:
Updated this branch to current `main`; current head is `aab8b9b3`.

This is an ordinary merge/update commit. Local validation: `cargo fmt --check`; `git diff --check 0144dad2..HEAD`. Hosted checks are running on the new head.

Komzpa:
Updated this branch to current `main`; current head is `22ab14f1`.

Komzpa:
Updated this branch to current `main` (`1d0098d`); current head is `176b3e6d`.

strk-ai-agent (commented):
Reviewed locally on `1d0098d` base. `cargo fmt --check`, `cargo check --locked`, and `git diff --check` all clean. The wiring is correct: `KeysCommand::export` → `InnerServer::export_keys` → `client.encryption().export_room_keys(file, &passphrase, predicate)` matches the SDK signature (`matrix-sdk-0.14.0/src/encryption/mod.rs:1402`), and `InboundGroupSession::room_id() -> &RoomId` is what the predicate receives (`matrix-sdk-crypto-0.14.0/src/olm/group_sessions/inbound.rs:463`). Issue #33 is satisfied as a first pass.

A few notes before merge:

**1. (Must-fix) Silent empty export when `--room-id` matches nothing.** `export_room_keys` returns `Result<(), Error>` with no count, so a successful export of zero matched sessions is indistinguishable from a full export. The user runs `/keys export --room-id !wrong:matrix.org file passphrase` and sees "Successfully exported E2EE keys" with an empty file. Suggest either:
  - Pre-flight count via the SDK store before calling `export_room_keys`, and print `Exported N keys for room X` or `No keys matched --room-id X; wrote empty file`.
  - At minimum, mention the `--room-id` value in the success message when one was given.
This is the most likely user-visible papercut.

**2. (Worth fixing) Case-sensitive comparison + port-included room IDs.** Read `ruma-identifiers-validation-0.11.0/src/room_id.rs::validate`: `RoomId::parse` checks sigil, max length, and NUL byte but does **not** lowercase. So:
  - User types `--room-id !Abc:Matrix.Org` and store has `!abc:matrix.org` → no match.
  - User types `--room-id !id:example.com:443` and store has `!id:example.com` → no match (port-included forms are accepted by the parser; see `ruma-common-0.16.0/src/identifiers/room_id.rs` tests).
Conservative (won't export wrong-room keys), but frustrating. Two options:
  - Lowercase the input in the clap validator before passing to `RoomId::parse`.
  - Compare with `eq_ignore_ascii_case` and strip an optional `:port`.

**3. (Optional, pre-existing) `export_keys` panics if there's no client.** `let client = self.get_client().unwrap();` (server.rs:1082) before the `if let Some(c) = self.connection()` guard (server.rs:1095). The guard never runs if the client was already `None` — the unwrap has already panicked. Since you're touching this function, consider moving the connection check up and dropping the `unwrap()` (e.g., `match self.get_client() { Some(c) => ..., None => self.print_error(...) }`). Not blocking, but it's right next door.

**4. (Optional) Predicate style.** `room_id.as_ref().map_or(true, |r| session.room_id().as_str() == r.as_str())` works but reads more directly as a `match`:
  ```rust
  |session| match &room_id {
      Some(r) => session.room_id().as_str() == r.as_str(),
      None => true,
  }
  ```
Pure style.

**5. (Note) Double-parse of room ID.** Clap validator calls `RoomId::parse` to validate, then `run` re-parses via `RoomId::parse(...).unwrap()`. clap 2.x doesn't let you stash a parsed value, so this is idiomatic — just noting the unwrap depends on the validator having run first (it has, by construction).

**6. (Note) No tests.** Same as the rest of the recent PRs; clap+WeeChat runtime is hard to mock. CI matrix passes on the head.

Verdict: **approve pending #1**. #2 is the next-most-likely user-visible issue.

Komzpa:
Updated to current `main` (`8f219a7`) at head `65ba097a` and addressed the review:

- filtered exports now name the exact room and explicitly say that the SDK cannot report whether any sessions matched;
- disconnected exports report an error instead of unwrapping a missing client;
- room IDs remain exact, opaque identifiers rather than being lowercased or otherwise normalized.

PR #146, opened 2026-06-23
build: run cargo before installing plugin

Komzpa:
`make install` now always runs the Cargo build target before copying the plugin, so a plain install does not skip Cargo's build-script dependency tracking.

The README notes that `WEECHAT_PLUGIN_FILE` is only used while building, and that rerunning `cargo build` or `make install` is enough after changing it.

The rust-weechat-side header input tracking is in poljar/rust-weechat#47.

Fix requested by @strk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,9 @@ that the `clang` command-line tools, C/C++ compiler headers, and matching
 `libclang` package are installed. On minimal distributions the library package
 alone may not provide the compiler include paths that `bindgen` needs.
 
+`WEECHAT_PLUGIN_FILE` is only used while building. After changing it, rerun
+`cargo build` or `make install`; `make clean` is not needed.
+
 After the dependencies are installed the plugin can be compiled with:
 
     cargo build --release
--- a/Makefile
+++ b/Makefile
@@ -6,17 +6,17 @@ SOURCES := $(wildcard src/*.rs src/bar_items/*.rs src/commands/*.rs src/room/*.r
 
 PROFILE ?= release
 
-.PHONY: install install-dir lint all help deb
+.PHONY: install install-dir lint all help deb FORCE
 
 all: help
 
 help: ## Print this help message
 	@grep -E '^[a-zA-Z._-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
-target/debug/libmatrix.so: $(SOURCES) ## Build plugin in dev profile
+target/debug/libmatrix.so: FORCE $(SOURCES) ## Build plugin in dev profile
 	cargo build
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -37,6 +37,9 @@ path to that header:
 
     export WEECHAT_PLUGIN_FILE=/usr/include/weechat/weechat-plugin.h
 
+`WEECHAT_PLUGIN_FILE` is only used while building. After changing it, rerun
+`cargo build` or `make install`; `make clean` is not needed.
```
strk, line 46: `make clean` is not even supported, unless it's buried in one of those many PRs ?
Unless supported, I'd rather not mention it here. 

This is one of those cases in which I'd appreciate a ./configure step, because that's usually the time in which header files are found, and requirements are checked. That way I can run ./configure once to automatically find the appropriate weechat-plugin.h (or force it using a --with-weechat-include <path>) and then be sure my configured build tree is setup to use that.

I've been told many times by @lnicola that I shouldn't insist in thinking "./configure && make" so before implementing my idea I'd rather hear from @poljar as the lead maintainer what his preference is.

I'm also happy to read from experienced Rust developers what the recommended way is (chances are there's a concept of "build profiles" in Cargo that we could be using for this).

Sorry for my doubts-introducing review, having doubts might be the only thing we humans are better at :)

lnicola: The convention is to support environment variables, but also try `pkgconf`. You've seen this before in the `openssl` docs: https://docs.rs/openssl/latest/openssl/#manual. At least on my distro, `weechat-devel` includes a `weechat.pc` file.

https://github.com/poljar/rust-weechat/blob/main/crates/weechat-sys/build.rs also has a bundled version of the bindings, but I don't know how often the WeeChat plugin ABI changes to have an opinion on whether it's useful.

lnicola:
Anyway, the title mentions Cargo, but it's only a `Makefile` change. From what I remember, having to run `make clean` or `distclean` isn't that unusual, so you can still do that without extra changes.

The actual tracking should happen in https://github.com/poljar/rust-weechat/blob/main/crates/weechat-sys/build.rs#L78. If the variable changes, the build script there should run again, causing `weechat-sys` and its reverse dependencies to be recompiled. So unless I'm missing something, Cargo should already handle it correctly.

README.md
Komzpa: Adjusted this PR to stay small: it only makes `make install` run Cargo before copying the plugin, and the README avoids saying `make clean` is needed.

For the rust-weechat side, #47 tracks the selected header inputs. I am not adding a configure-style layer here unless @poljar wants that direction.

Komzpa:
Closing this one as obsolete after lnicola's comment.

The actual Cargo-side tracking landed in poljar/rust-weechat#47, and `poljar/rust-weechat` `main` now has `bindgen::CargoCallbacks::new()` in `crates/weechat-sys/build.rs` plus `rerun-if-env-changed=WEECHAT_PLUGIN_FILE`.

So this Makefile workaround is no longer the right fix here. Thanks for pointing at the correct layer.

PR #148, opened 2026-06-23
fix: fetch history for empty room buffers

### LINKED ISSUE #55: History of chat is not fetched
Upon re-connection,  previous session's chat log is not shown. I think there is no need to "save" any logs upon exit since Matrix protocol saves the logs in server-side.

Thanks for this project 😊

Komzpa:
When switching to an empty Matrix room buffer, fetch the first available history batch automatically instead of waiting for a manual PageUp.

The fetch stays conservative: it only runs for empty buffers, and the existing in-flight guard prevents duplicate history requests. This only auto-triggers the existing history path.

Fixes #55.

Fix requested by @strk.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -167,12 +167,21 @@ impl SignalCallback for Servers {
     fn callback(
         &mut self,
         _: &Weechat,
-        _signal_name: &str,
+        signal_name: &str,
         data: Option<SignalData>,
     ) -> ReturnCode {
         if let Some(SignalData::Buffer(buffer)) = data {
             if let Some(room) = self.find_room(&buffer) {
-                room.update_typing_notice();
+                match signal_name {
+                    "buffer_switch" => {
+                        Weechat::spawn(async move {
+                            room.get_messages_if_empty().await
+                        })
+                        .detach();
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -167,12 +167,21 @@ impl SignalCallback for Servers {
     fn callback(
         &mut self,
         _: &Weechat,
-        _signal_name: &str,
+        signal_name: &str,
```
strk, line 170: Could you explain the callback name change ? Was it really needed ? Why so ?

Komzpa: Yes, this is needed because the same `Servers` signal callback is now registered for both `input_text_changed` and `buffer_switch`.

Before this PR the callback only handled typing notices, so the signal name could be ignored as `_signal_name`. This PR adds the `buffer_switch` hook using the same callback object, so it has to inspect `signal_name` to dispatch:

- `input_text_changed` keeps updating the typing notice.
- `buffer_switch` fetches history only when the switched-to room buffer is empty.

Without the rename/use of `signal_name`, the callback could not tell those two signals apart.

Issue #31, opened 2021-01-03
Display the verification state of the room in the status bar.


poljar:
This first needs support in the rust-sdk, after that it should be as simple as checking if the room is encrypted and showing a sign for the different verification states.

strk:
This seems to be done already, anything left before closing this ticket, @poljar ?

PR #152, opened 2026-06-23
fix: refresh room verification status

Komzpa:
Refreshes the room status bar when room state, membership, or verification events can change the signs shown in `buffer_modes`.

The status item already renders encrypted rooms and the unverified-device warning. This makes the item repaint when the state behind those signs changes instead of waiting for an unrelated refresh.

References #31.

Fix requested by @strk.

### DIFF
--- a/src/room/mod.rs
+++ b/src/room/mod.rs
@@ -392,6 +392,10 @@ impl BufferInputCallbackAsync for MatrixRoom {
 }
 
 impl MatrixRoom {
+    fn update_status_bar(&self) {
+        Weechat::bar_item_update("buffer_modes");
+    }
+
     pub fn is_encrypted(&self) -> bool {
         self.members
             .runtime
@@ -972,7 +976,9 @@ impl MatrixRoom {
                 ambiguity_change,
                 smart_filter_delay_ms,
             )
-            .await
+            .await;
+
+        self.update_status_bar();
     }
 
     fn set_prev_batch(&self) {
@@ -1054,5 +1060,7 @@ impl MatrixRoom {
             AnySyncStateEvent::RoomCanonicalAlias(_) => self.buffer.set_alias(),
             _ => (),
         }
+
+        self.update_status_bar();
     }
 }
--- a/src/room/mod.rs
+++ b/src/room/mod.rs
@@ -417,10 +417,6 @@ impl BufferInputCallbackAsync for MatrixRoom {
 }
 
 impl MatrixRoom {
-    fn update_status_bar(&self) {
-        Weechat::bar_item_update("buffer_modes");
-    }
-
     pub fn is_encrypted(&self) -> bool {
         self.members
             .runtime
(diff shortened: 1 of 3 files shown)


Komzpa:
Updated this branch to current `main`; current head is `31df71ed`.

This is an ordinary merge/update commit. Local validation: `cargo fmt --check`; `git diff --check 953c2d6c..HEAD`. Hosted checks are running on the new head.

Komzpa:
Updated this branch to current `main`; current head is `d451d9de`.

Komzpa:
Updated this branch to current `main` (`1d0098d`); current head is `f04dd79c`.

strk-ai-agent (commented):
Reviewed locally on `1d0098d` base. `cargo fmt --check`, `cargo check --locked`, and `git diff --check` all clean. The four refresh sites correctly cover every transition that can flip a `buffer_modes` sign (encryption, devices, public, busy).

A few notes before merge:

**1. Helper / direct-call style is mixed.** `src/room/mod.rs` introduces `MatrixRoom::update_status_bar`, but `src/room/verification.rs` and `src/server.rs` call `Weechat::bar_item_update("buffer_modes")` directly. Either propagate the helper to all four sites (a free fn in a `bar_items` module, or a small `BarItems` struct held by `InnerServer`/`MatrixRoom`), or drop the helper and call `bar_item_update` directly in `room/mod.rs` too. The inconsistency is the worst of both.

**2. `handle_sync_state_event` refresh is overly broad.** The new `update_status_bar()` after the `match event` runs for *every* state event, including `RoomName`, `RoomTopic`, `RoomCanonicalAlias`, `SpaceParent` — none of which change any of the four signs. Cheap, but conceptually wrong. A surgical version would add `RoomEncryption(_)` and `RoomJoinRules(_)` arms that set a local `refresh = true`, and only update the bar item if that flag was set. Optional polish.

**3. `MatrixRoom::update_status_bar(&self)` doesn't use `&self`.** The receiver is a stub. If you keep the helper, fine; if you refactor per #1, prefer a free function or a method on a struct where it has a real reason to be a method.

**4. `matrix_modes` inconsistency (pre-existing).** `get_messages` (lines 911/912 and 944/945 of `src/room/mod.rs`) calls both `buffer_modes` and `matrix_modes`, but there is no `BarItem::new(\"matrix_modes\", ...)` anywhere in this repo — the `matrix_modes` calls are no-ops. PR doesn't make it worse, but doesn't fix it either. Worth a follow-up to remove the dead calls.

**5. `receive_to_device_event` covers the right events.** Confirmed that `KeyVerificationMac` is the hook that flips `sas.is_done()` (see `src/verification_buffer.rs:330-338`), which is what the SDK uses to mark devices as verified. No need to add `KeyVerificationDone` — `Mac` is sufficient.

**6. No tests.** Understandable for a WeeChat-bar-item change; CI matrix is running on the head.

Verdict: **approve** once #1 is harmonized. #2 is optional polish.

Komzpa:
Updated to current `main` (`8f219a7`) at head `968c3925` and addressed the review:

- status-bar refresh uses one direct call style;
- state events refresh only for encryption and join-rule changes;
- to-device verification completion now refreshes the status item and reaches the existing event handler.

PR #154, opened 2026-06-23
feat: support SSO login

### LINKED ISSUE #28: Support login with SSO.
The Python script supports SSO login by starting a local HTTP server that waits for the loginToken. The helper for the Python script can be found over [here](https://github.com/poljar/weechat-matrix/blob/master/contrib/matrix_sso_helper.py).

We'll likely want an in-process server instead.

### LINKED ISSUE #108: Feature Request: Single sign-on support
Feature Request:  
please support Single sign-on like the python version   
[https://github.com/poljar/weechat-matrix?tab=readme-ov-file#single-sign-on](https://github.com/poljar/weechat-matrix?tab=readme-ov-file#single-sign-on)

Komzpa:
Adds first-login SSO support using the Matrix SDK's built-in `sso-login` flow.

If the server password option is empty, `/matrix connect` checks whether the homeserver advertises SSO, prints the browser login URL in WeeChat, and lets the SDK's local callback server finish the login. Normal username/password login is unchanged when a password is configured.

The branch is split into two commits: first enabling the Matrix SDK `sso-login` feature and lockfile changes, then the weechat-matrix-rs login flow and README note.

This keeps the scope smaller than poljar/weechat-matrix-rs#119: no manual token command, no extra config switch, and no custom callback server in this plugin. Manual token entry is tracked separately in poljar/weechat-matrix-rs#172.

Fixes poljar/weechat-matrix-rs#28.
Fixes poljar/weechat-matrix-rs#108.
References poljar/weechat-matrix-rs#119.

Checked with `cargo fmt --check`, `git diff --check`, and Docker `WEECHAT_BUNDLED=1 CARGO_TARGET_DIR=/target cargo check --locked`.

Fix requested by @strk.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,6 +238,59 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
+name = "axum"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "021e862c184ae977658b36c4500f7feac3221ca5da43e3f25bd04ab6c79a29b5"
+dependencies = [
+ "axum-core",
+ "bytes",
+ "form_urlencoded",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
+ "itoa",
+ "matchit",
+ "memchr",
+ "mime",
+ "percent-encoding",
+ "pin-project-lite",
+ "rustversion",
+ "serde",
+ "serde_json",
+ "serde_path_to_error",
+ "serde_urlencoded",
+ "sync_wrapper",
+ "tokio",
+ "tower",
+ "tower-layer",
+ "tower-service",
+ "tracing",
+]
+
+[[package]]
+name = "axum-core"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08c78f31d7b1291f7ee735c1c6780ccde7785daae9a9206026862dab7d8792d1"
+dependencies = [
+ "bytes",
+ "futures-core",
+ "http",
+ "http-body",
+ "http-body-util",
+ "mime",
+ "pin-project-lite",
+ "sync_wrapper",
+ "tower-layer",
+ "tower-service",
+ "tracing",
+]
+
+[[package]]
 name = "backon"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1466,6 +1519,7 @@ dependencies = [
  "http",
  "http-body",
  "httparse",
+ "httpdate",
  "itoa",
  "pin-project-lite",
  "pin-utils",
@@ -2051,6 +2105,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "matchit"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e1ffaa40ddd1f3ed91f717a33c8c0ee23fff369e3aa8772b9605cc1d22f4c3"
+
+[[package]]
 name = "matrix-pickle"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2085,6 +2145,7 @@ dependencies = [
  "async-channel",
  "async-stream",
  "async-trait",
+ "axum",
  "backon",
  "bytes",
  "bytesize",
@@ -2112,6 +2173,7 @@ dependencies = [
  "once_cell",
  "percent-encoding",
  "pin-project-lite",
+ "rand 0.8.5",
  "reqwest",
  "ruma",
  "serde",
@@ -2123,6 +2185,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tokio-util",
+ "tower",
  "tracing",
  "url",
  "urlencoding",
@@ -4191,6 +4254,7 @@ dependencies = [
  "tokio",
  "tower-layer",
  "tower-service",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


strk:
I love the idea of finally being able to NOT write passwords in the config file ! Did you try this against the OSGeo matrix server ?

As this PR includes a migration to a newer matrix-sdk, would it be cleaner to do that in its own commit rather than mixing functional and "structural" changes ? Would that be hard for you to do now ?

strk:
@bogdando your review here is very welcome, being the author of the other version of the feature. I suspect the ability of manually entering tokens is important (things happen), could it easily be added in a second step ?

Komzpa:
Split now: first commit is only the Matrix SDK `sso-login` feature/lockfile, second commit is the plugin SSO flow and README note.

I have not tested against the OSGeo Matrix server specifically; validation so far is compile-level. Manual token entry sounds useful, but I would keep it as a follow-up after this SDK SSO path lands.

Komzpa:
Rebased onto current `main` after the recent README/doc merges. The conflict is gone; the README keeps both the SSO note and the first-connect patience note.

`cargo fmt --check`, `git diff --check upstream/main..HEAD`, and Ubuntu 24.04 Docker/rustup `WEECHAT_BUNDLED=1 cargo check --locked` pass locally.

Komzpa:
Follow-up cleanup for the SSO review thread:

- the Matrix SDK feature/lockfile change is already split from the plugin SSO flow;
- I still have not tested this against the OSGeo homeserver specifically, so the PR body keeps the validation claim at compile/check level;
- the manual token fallback is now tracked separately in poljar/weechat-matrix-rs#172 instead of being a loose "later" promise on this PR.

Komzpa:
Updated this branch to current `main`; current head is `6ee45766`.

This is an ordinary merge/update commit. Local validation: `cargo fmt --check`; `git diff --check 32bbe601..HEAD`. Hosted checks are running on the new head.

Komzpa:
Updated this branch to current `main`; current head is `d5c453a7`.

Komzpa:
Updated this branch to current `main` (`1d0098d`); current head is `b280aa7b`.

strk-ai-agent (commented):
Reviewed locally on `1d0098d` base. `cargo fmt --check`, `cargo check --locked`, and `git diff --check` all clean. Wiring is correct: empty password → `get_login_types()` → check `LoginType::Sso(_)` → `SsoLoginBuilder::send().await` blocking on the SDK's local server (`matrix-sdk-0.14.0/src/utils/local_server.rs:96-115`, default port range `20000..30000` on `LocalhostAny`) → existing `receive_login` path on success. `matrix-sdk`'s `sso-login` feature is enabled in `Cargo.toml`, with `axum`/`axum-core`/`matchit` as new transitive deps. Scope is appropriately minimal vs. the abandoned #119.

A few notes:

**1. (Style) Two-arm if/else with duplicated builder plumbing.** The SSO and password branches each repeat the same `.initial_device_display_name("WeeChat-Matrix-rs")` + conditional `.device_id(...)` setup. The two builder types are unrelated, so you can't trivially unify them, but the trailing `device_id` and the display-name line are identical. Consider a small helper closure or trait that applies both — or just a comment that the duplication is forced by the SDK API.

**2. (Worth fixing) Redundant clone in the SSO callback.**
```rust
let sso_channel = channel.clone();
let mut builder = matrix_auth
    .login_sso(move |sso_url| {
        let sso_channel = sso_channel.clone();   // <-- redundant
        async move { ... }
    })
```
The outer `move |sso_url|` already moves `sso_channel` into the closure; the async block can capture it by move from there. Drop the inner clone. Also the name `sso_channel` shadows the outer `channel` parameter inside the closure — clearer names would help.

**3. (Minor) `Ok(())` semantics in the SSO callback.** SDK docs say the callback "should usually be used to open the SSO URL in a browser and must return `Ok(())` if the URL was successfully opened". This plugin doesn't open the URL — it hands it to WeeChat to print, and the user opens it themselves. Returning `Ok(())` always is the right call (any `Err` would abort the entire login flow, and we *did* successfully hand the URL off), but worth a brief source comment so a future reader doesn't think "we should `Err` on channel failure".

Side note: `let _ = sso_channel.send(...).await;` silently drops the channel-send error. Consistent with the existing sync_loop send patterns, but if WeeChat has died the URL never reaches the user and the SDK still waits on the callback server with no timeout.

**4. (Minor) README doesn't mention the homeserver SSO requirement.** New line says "WeeChat will print a login URL when you connect" but doesn't say "if your homeserver advertises SSO". The error path "No password configured and homeserver does not advertise SSO login" is good, but the README should hint at this so users on non-SSO homeservers aren't confused when nothing happens.

**5. (Note) Device-id file rename is correct.** `<username>.device_id` → `<server_name>.device_id` when `username` is empty. Non-empty usernames keep the existing filename. `save_device_id` is updated symmetrically, so save/load are consistent. The crypto store is unchanged (still keyed on `get_server_path()` and `get_server_cache_path()` with `DEFAULT_PASSPHRASE`), so reusing the device_id across SSO logins will recover the encryption keys as required by the SDK's `SsoLoginBuilder::device_id` contract.

**6. (Note) SDK port range.** Defaults to `20000..30000` on `LocalhostAny`. Paranoid localhost firewalls on that range would silently block SSO. Document or expose a `LocalServerBuilder` if it ever comes up.

**7. (Note) `identity_provider_id` not exposed.** `SsoLoginBuilder::identity_provider_id(...)` exists for homeservers with multiple IdPs. Default ("any") is fine for v1; worth a follow-up if anyone asks.

**8. (Note) No tests.** Same as the rest of the recent PRs; SSO can't be unit-tested without a real homeserver. CI matrix passes on the head.

Verdict: **approve**. Items #1 and #2 are the only style nits I'd actually want addressed; #3 and #4 are documentation polish.

Komzpa:
Updated to current `main` (`8f219a7`) at head `aa84805b` and addressed the review:

- removed the redundant callback sender clone and documented why handing the URL to WeeChat completes the callback;
- clarified that the homeserver must advertise SSO;
- fixed the callback port at `127.0.0.1:29325` and documented SSH forwarding for remote WeeChat hosts.

Komzpa:
Rebased onto current `upstream/main` (`0f04ee46deae5a67618b57fb844486f100715e4d`). Current head is `2c7e3ecc4bbdccc4c4f6e5c3e69c57629bf6c19b`; all four hosted jobs pass: [checks](https://github.com/poljar/weechat-matrix-rs/pull/154/checks). Ready for rereview.

PR #155, opened 2026-06-23
fix: split Matrix SDK event cache storage

### LINKED ISSUE #70: Using matrix-rs somtimes generate a lot of disk-I/O
When i load the plugin and it goes online weechat starts grinding disk like i've never seen. I get constant 600kB-1M/s disk access for days and days. As soon as i unload / offline the plugin the activity stops. This happens every now and then.... What's this about?

Komzpa:
Separates the Matrix SDK event cache from the state and crypto stores.

The existing per-server directory keeps Matrix state and encryption data. The event cache now goes into a sibling `<server>-cache` directory, so cache cleanup no longer requires deleting the important state directory.

The README documents which directory is safe to remove when cache churn or disk I/O becomes a problem.

This is a narrow mitigation for the disk-I/O report, not a claim that every possible write source is gone.

Fixes #70.

Checked with `cargo fmt --check`, `git diff --check`, and Docker `WEECHAT_BUNDLED=1 CARGO_TARGET_DIR=/tmp/weechat-cache-split-check cargo check --locked`.

Fix requested by @strk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,15 @@ The inactivity delay defaults to five minutes and can be changed with:
 
        /set matrix-rust.look.smart_filter_delay 300000
 
+# Storage
+
+The plugin keeps Matrix state and encryption data in
+`$WEECHAT_HOME/matrix-rust/<server>/`. It keeps the Matrix SDK event cache in
+`$WEECHAT_HOME/matrix-rust/<server>-cache/`.
+
+If the cache grows or causes too much disk I/O, disconnect WeeChat and remove
+only the `-cache` directory. Do not remove the main server directory unless you
+want to reset stored Matrix state.
 
 # Helpful Commands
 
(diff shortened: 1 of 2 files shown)


strk:
@aiobofh can you try this branch ?

strk:
@Komzpa what's in the cache ? Is the backlog there ? Does upgrading to this branch migrate old caches to new location or will it have to re-create the full cache ? I ask because of #183 (I might have tested this branch at some point, on the slow machine)

Komzpa:
The new `<server>-cache` directory is the Matrix SDK event cache. It can include cached timeline/backlog data, so removing it is safe in the sense that the client can rebuild it, but it may need to re-fetch/rebuild that cache afterward.

#155 only changes the cache path; it does not migrate an old event cache into the new sibling directory. The important state/crypto data stays in `$WEECHAT_HOME/matrix-rust/<server>/`, and that main directory should not be removed for cache cleanup.

Issue #95, opened 2025-03-19
Authenticated media support


strk:
Newer Matrix releases resulted in impossibility to access uploaded media without sending an authentication header.

https://matrix.org/blog/2024/06/20/matrix-v1.11-release/

This breaks the weechat way to access those resources, which relied on a simple url.

I guess it would take some matrix-aware script to grab the url and send an authenticated request ?

The python version is also affected: https://github.com/poljar/weechat-matrix/issues/369

strk:
This was added with GH-156 - improvements are worth a new ticket

PR #156, opened 2026-06-23
feat: add authenticated media download command

Komzpa:
Adds a plain Matrix media download path through the logged-in Matrix client.

`/matrix media download <mxc-uri> <file>` asks matrix-sdk to fetch the MXC URI, so homeservers that require authenticated media can still be used for plain media. The command refuses to overwrite an existing file.

Plain media messages also print the exact `/matrix media download ... <file>` command as a hint. Encrypted media keeps the existing `emxc://` rendering; supporting authenticated encrypted downloads needs the event's encrypted-file metadata and is intentionally outside this small batch.

References #95.

Checked with:

- `cargo fmt --check`
- `git diff --check upstream/main..HEAD`
- Ubuntu 24.04 Docker with Rust 1.96.0, `WEECHAT_BUNDLED=1`, and `CARGO_TARGET_DIR=/tmp/weechat-pr156-test cargo test --locked media_download_command --lib`

Fix requested by @strk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,3 +122,5 @@ _(replace the placeholders in brackets [] with your own details)_:
 
 Room buffers accept normal message input after `/matrix connect [server-name]`
 has completed.
+
+`/matrix media download <mxc-uri> <file>` downloads Matrix media through the logged-in client, including homeservers that require authenticated media. It refuses to overwrite an existing file.
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -12,6 +12,7 @@ mod devices;
 mod keys;
 mod matrix;
 mod me;
+mod media;
 mod page_up;
 mod verification;
 
@@ -20,6 +21,7 @@ use devices::DevicesCommand;
 use keys::KeysCommand;
 use matrix::MatrixCommand;
 use me::MeCommand;
+use media::MediaCommand;
 use page_up::PageUpCommand;
 
 pub struct Commands {
--- a/src/commands/matrix.rs
+++ b/src/commands/matrix.rs
@@ -12,7 +12,7 @@ use weechat::{
 
 use super::{parse_and_run, verification::VerificationCommand};
 use crate::{
-    commands::{DevicesCommand, KeysCommand},
+    commands::{DevicesCommand, KeysCommand, MediaCommand},
     config::ConfigHandle,
     MatrixServer, Servers, PLUGIN_NAME,
 };
@@ -34,6 +34,7 @@ impl MatrixCommand {
             .add_argument("connect <server-name>")
             .add_argument("devices delete|list|set-name")
             .add_argument("keys import|export <file> <passphrase>")
+            .add_argument("media download <mxc-uri> <file>")
             .add_argument("disconnect <server-name>")
             .add_argument("reconnect <server-name>")
             .add_argument("help <matrix-command> [<matrix-subcommand>]")
@@ -44,16 +45,19 @@ impl MatrixCommand {
    reconnect: Reconnect to server(s).
      devices: {}
         keys: {}
+       media: {}
 verification: {}
         help: Show detailed command help.\n
 Use /matrix [command] help to find out more.\n",
                 DevicesCommand::DESCRIPTION,
                 KeysCommand::DESCRIPTION,
+                MediaCommand::DESCRIPTION,
                 VerificationCommand::DESCRIPTION,
             ))
             .add_completion("server add|delete|list|listfull")
             .add_completion("devices list|delete|set-name %(matrix-users)")
             .add_completion(format!("keys {}", KeysCommand::COMPLETION))
+            .add_completion(format!("media {}", MediaCommand::COMPLETION))
             .add_completion(format!(
                 "verification {}",
                 VerificationCommand::COMPLETION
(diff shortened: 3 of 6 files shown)


strk:
conflicts (the problem of too many PRs, I know..)

README.md
```diff
@@ -102,3 +102,5 @@ Configuration is completed primarily through the Weechat interface. First start
 `/help matrix` will print information about the `/matrix` command.
 
 `/matrix help [command]` will print information for subcommands, such as `/matrix help server`
+
+`/matrix media download <mxc-uri> <file>` downloads Matrix media through the logged-in client, including homeservers that require authenticated media. It refuses to overwrite an existing file.
```
strk: Does <mxc-uri> mean the uri MUST start with mxc: ? Does this change ensure the attachment URLs are printed with that schema ? Because right now what I see in the buffer are lines like this:

>  <image.png>[¹https://matrix.osgeo.org/_matrix/media/r0/download/matrix.org/nNLgLbzLWXcaGkbDRcWEHOPY] 

The 1 subscript is inserted by an "url_hint.py" script that lets me do things like `/open_url 1` invoking an external command to deal with the url. Ideally I'd keep the same convention so I'd need the url passed to the external command to have the mxc schema to know what to do with it, and probably a way for the external command to interact with the running weechat matrix plugin to know how to authenticate.

I'm sure LLMs can cope with reviewing this comment of mine and suggest a solution, shall we try that part too ? (open source project management is not just code fixes but also conversation).

Komzpa:
Rebased onto current `main`; the conflict should be gone now. CI is running on the new head.

README.md
Komzpa: Yes: the command accepts Matrix `mxc://...` URIs, not homeserver HTTP download URLs.

This branch now also prints a second line for plain media:

`authenticated download: /matrix media download mxc://... <file>`

So an external URL helper can see/copy the original MXC URI, while the actual authenticated fetch stays inside the logged-in WeeChat plugin. A more direct `/open_url` integration sounds useful, but I would keep that as a follow-up after this lands.

Komzpa: Done in `c48c4dd`.

Plain media now prints an `mxc://` line. `/open_url` and `/input open_url` catch those Matrix media URLs and download through the logged-in client; non-Matrix URLs pass through.

fix requested by @strk

src/commands/mod.rs
```diff
@@ -45,6 +51,11 @@ impl Commands {
             _verification: VerificationCommand::create(servers)?,
             _buffer_clear: BufferClearCommand::create(servers)?,
             _me: MeCommand::create(servers)?,
+            _open_url: OpenUrlCommand::create(servers, "/open_url *")?,
```
strk: Sorry I wasn't clear what I meant by `/open_url`, I was referring to an alias I defined on my system, is this change here introducing another command ? Is it `/matrix open_url` ?

strk (dismissed):
I don't understand the `/open_url` and `/input open_url` parts of this PR, maybe it bloated out of scope ? You mentioned  you was going to do that part as a follow-up ? I'm not sure what you are doing but I'd rather not hook on "core commands" if that's what is tried here, beacuse `/open_url` (that I'm using) is just an alias that's suggested by the help text of a `url_hint.py` script, nothing core, just something that I happen to have defined myself...

strk:
What I get BEFORE applying this patch:
image

(will build and report what I see afterwards)

Komzpa:
The current head drops the `/open_url` / `/input open_url` hook entirely. This PR is back to the plain `/matrix media download <mxc-uri> <file>` path plus the rendered hint.

strk:
Lovely, I gave it a try and IT WORKED !

My personal review:

 1. I'd like to see a feedback when I call `/matrix media download` (did it succeed? Where is the file?)
 2. I'd like to be able to NOT specify a target filename but have some default
 3. I'd like TAB-completion: /matrix media download <TAB>


This is what I see when "strk" uploads a file from FluffyChat/web:
```
<&strk (osgeo)> <The moon and I>[https://matrix.osgeo.org/_matrix/media/r0/download/osgeo.org/HeoksrELYbTnezttjrmDZuWx]
<&strk (osgeo)> authenticated download: /matrix media download mxc://osgeo.org/HeoksrELYbTnezttjrmDZuWx <file>
```

It would make more sense to see something that feels like:
```
< *> strk (osgeo) attached `The moon and I`: https://matrix.osgeo.org/_matrix/media/r0/download/osgeo.org/HeoksrELYbTnezttjrmDZuWx
<--> Notice(matrix): authenticated download: /matrix media download mxc://osgeo.org/ODUmhlLHFBiSTsrSzfWhoJWF [<file>]
```

I'm still unsure if the https:// line is even useful, anyway my next line would be (if it looks interesting):
```
/matrix media download <TAB>
```

Which I expect to be completed to:
```
/matrix media download mxc://osgeo.org/ODUmhlLHFBiSTsrSzfWhoJWF
```

Then I'd expect a message like:
```
Media was saved to <path>
```

strk:
@Komzpa let me know if you prefer to have this merged and then make the improvements as a follow-up, I'm fine either way

strk:
Actually, I think I'm ok with merging this, and refine later. Better have a way to download media than nothing :)

PR #161, opened 2026-06-29
Add Debian packaging and 'deb' Makefile target

strk-ai-agent:
### Summary

This PR adds Debian packaging support to the project, allowing users to build a `.deb` package for easy installation on Debian/Ubuntu systems.

### Changes

- **New branch:** `debian`
- **`debian/` directory** with standard Debian packaging files:
  - `changelog` — version 0.1.0-1 (UNRELEASED)
  - `control` — package metadata with build/runtime dependencies (cargo, rustc, debhelper, weechat-plugins)
  - `copyright` — ISC license information
  - `rules` — build using `dh` with overrides for Rust (`cargo build --release`) and install (plugin `.so` → `/usr/lib/weechat/plugins/matrix.so`)
  - `compat` — debhelper 13
  - `source/format` — 3.0 (native)
- **`Makefile`** — added `deb` target that runs `dpkg-buildpackage -b --no-sign`

### Usage

```bash
make deb          # builds the .deb in the parent directory
sudo dpkg -i ../weechat-matrix_0.1.0-1_amd64.deb  # install
```

### Notes

- The plugin installs to `/usr/lib/weechat/plugins/matrix.so`, the standard WeeChat system plugin directory
- Requires `dpkg-dev`, `cargo`, `rustc`, and `debhelper` to build

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,21 @@ authors = ["Damir Jelić <poljar@termina.org.uk>"]
 edition = "2018"
 license = "ISC"
 resolver = "2"
+description = "WeeChat plugin for the Matrix protocol"
+
+[package.metadata.deb]
+maintainer = "Damir Jelić <poljar@termina.org.uk>"
+copyright = "2019-2024, Damir Jelić <poljar@termina.org.uk>"
+extended-description = """\
+WeeChat plugin for the Matrix protocol built with the Matrix Rust SDK.
+Provides native Matrix support within WeeChat, allowing you to chat
+on Matrix from your favourite IRC client."""
+section = "net"
+priority = "optional"
+depends = "$auto, weechat-plugins"
+assets = [
+    ["target/release/libmatrix.so", "usr/lib/weechat/plugins/matrix.so", "755"],
+]
 
 [lib]
 name = "matrix"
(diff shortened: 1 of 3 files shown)


strk:
@poljar what do you think about this ? In order to test the software I'm currently building it on the server from which I use it, but that machine is slow at building... Let me know if you think it's ok to have a debian/ dir in the official repo, of so I'll ask my minion to document what needs be

@Komzpa your input (and your minion's one) is welcome too

Komzpa:
My take: keep this separate until @poljar says whether `debian/` belongs in the upstream repo.

If yes, I would rebase this as a focused packaging PR and require one real `dpkg-buildpackage -b --no-sign` run before merge. The current direction looks fine as a starting point, but the package metadata/build output should be checked against the actual artifact names and plugin path after that build.

If no, the useful part is still the documented build recipe; that could live in README or release notes without carrying a Debian packaging directory upstream.

strk-ai-agent:
cc @lnicola — could you take a look? Thanks!

strk:
> My take: keep this separate until @poljar says whether `debian/` belongs in the upstream repo.

I've refactored it to use `cargo-deb`, now there's no need for a debian/ dir

strk:
Superceeded by #163

PR #171, opened 2026-06-30
feat: improve media download UX

### LINKED ISSUE #168: Improve authenticated media download experience
See https://github.com/poljar/weechat-matrix-rs/pull/156#issuecomment-4840720972

Komzpa:
Fixes #168.

Improves the authenticated media download flow:

- `/matrix media download <mxc-uri>` now defaults the output filename to the MXC media id.
- `/matrix media download <TAB>` completes MXC URIs seen in the current Matrix buffer.
- The rendered helper line now uses WeeChat's network prefix, so it reads as a local hint rather than another user message.
- Existing download start/success/error feedback is kept.

Validation:
- `cargo fmt --check`
- `git diff --check`
- Docker: `WEECHAT_BUNDLED=1 CARGO_TARGET_DIR=/target cargo test --locked media --lib`
- Docker: `WEECHAT_BUNDLED=1 CARGO_TARGET_DIR=/target cargo test --locked extract --lib`

### DIFF
--- a/src/commands/media.rs
+++ b/src/commands/media.rs
@@ -7,7 +7,7 @@ use clap::{
 use matrix_sdk::ruma::MxcUri;
 use weechat::{buffer::Buffer, Weechat};
 
-use crate::Servers;
+use crate::{BufferOwner, Servers};
 
 pub struct MediaCommand;
 
@@ -16,9 +16,18 @@ impl MediaCommand {
     pub const COMPLETION: &'static str = "download %(matrix-media)";
 
     pub fn run(buffer: &Buffer, servers: &Servers, args: &ArgMatches) {
-        let server = match servers.find_server(buffer) {
-            Some(server) => server,
-            None => {
+        let (server, output_buffer) = match servers.buffer_owner(buffer) {
+            BufferOwner::Room(server, room) => {
+                (server, Some(room.buffer_handle()))
+            }
+            BufferOwner::Server(server) => {
+                let output_buffer = server.server_buffer().as_ref().cloned();
+                (server, output_buffer)
+            }
+            BufferOwner::Verification(server, verification) => {
+                (server, Some(verification.buffer()))
+            }
+            BufferOwner::None => {
                 Weechat::print("Must be executed on Matrix buffer");
                 return;
             }
@@ -51,7 +60,9 @@ impl MediaCommand {
                 };
 
                 Weechat::spawn(async move {
-                    server.download_media(uri.into(), file).await;
+                    server
+                        .download_media(uri.into(), file, output_buffer)
+                        .await;
                 })
                 .detach();
             }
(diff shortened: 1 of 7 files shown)


strk:
The TAB-completion works great, but:

 1. The feedback is given in the server buffer instead of the buffer from which I'm typing (annoying)
 2. The feedback doesn't include full path:  `matrix: Successfully downloaded media to VGJOEZj168GmoyNmibuntt6BcN6vjpIV` (where is it?)

strk:
Note that if you force-push w/out any comment I won't receive a notification so won't know it's time for me to review again -- I'm building the new version, will review (or directly merge) when ready

strk:
@Komzpa am I right that [8a6d936](https://github.com/poljar/weechat-matrix-rs/commit/8a6d9363367b10e116c007ba990d01289b013046) addresses none of the comments in https://github.com/poljar/weechat-matrix-rs/pull/171#issuecomment-4859449785 ? Do your minions require an official "review" to be triggered or should the above comment have done ?

Let me know if you prefer I just merge this as-is so to continue later (#168 is still not complete)

src/commands/media.rs
```diff
@@ -72,3 +81,27 @@ impl MediaCommand {
...
     ];
 }
+
+fn default_download_file(uri: &MxcUri) -> Option<PathBuf> {
+    uri.as_str()
+        .trim_end_matches('/')
+        .rsplit('/')
+        .next()
+        .filter(|name| !name.is_empty())
+        .map(PathBuf::from)
```
strk, line 163: It would be good to have a prefix used, otherwise I easily end up with lots of files with obscure names and don't know where they come from. How about starting with a `matrix-media-` prefix and making that configurable ? A configurable prefix could also include a path, to save media by default in some weechat-matrix state dir (like `~/.local/state/weechat-matrix`, but honouring XDG env).

strk (changes requested):
The TAB-completion works great, but:

    - The feedback should be given in the same buffer from which the `/matrix media download` command was given
    - The feedback should include the full path to the downloaded file (it's currently relative)

Komzpa:
Sorry, this should have been commented when I pushed the follow-ups.

Current head is `065f65ea3b3672d2d03325f633581c4cf0dcd552`. I added normal follow-up commits rather than force-pushing:

- `fdcc8ae` reports media download progress/errors/success in the buffer where the command was run, and success/error text now includes the full display path.
- `065f65e` adds the default `matrix-media-` prefix plus configurable `matrix-rust.media.download_prefix`; the prefix can include a directory and supports `~`, `$XDG_STATE_HOME`, and `${XDG_STATE_HOME}`.

GitHub Actions are green on this head.

strk (approved):
Sounds good to me, thank you !

PR #174, opened 2026-07-02
fix: avoid member-list buffer names

### LINKED ISSUE #173: Sometimes room titles are not fetched and they are replaced with list of members
Sometimes room titles/names not fetched and are replaced with list of members instead, look at the screenshot:

Image

This is the SAME version of weechat with the SAME version of the weechat-matrix-rs plugin that is also run here:

Image

See how the very long room titles negatively affect display of the buffer list, that end up being all in a single column and thus impossible to read.

I suspect it's a timeout becuase I know the osgeo.org matrix server is very slow and it seems to happen only against that server (of 3 servers I'm connected to)

\cc @Komzpa a job for your minions ?

Komzpa:
## Summary
- use explicit Matrix room names for WeeChat buffer short names
- fall back to canonical aliases, then room ids, instead of SDK-calculated member lists
- keep existing prefix behavior for channels, DMs, and room id fallbacks

## Testing
- cargo fmt --check
- git diff --check
- WEECHAT_BUNDLED=1 CARGO_TARGET_DIR=/work/target/docker-issue173 cargo test --locked preserves_ in rust:1.88-bookworm

Fixes #173

### DIFF
--- a/src/room/buffer.rs
+++ b/src/room/buffer.rs
@@ -2,13 +2,10 @@ use std::{borrow::Cow, cell::RefCell, rc::Rc};
 
 use matrix_sdk::{
     ruma::{EventId, OwnedRoomAliasId, TransactionId, UserId},
-    Room, StoreError,
+    Room,
 };
 use tokio::runtime::Handle;
-use weechat::{
-    buffer::{Buffer, BufferHandle, BufferLine, LineData},
-    Prefix, Weechat,
-};
+use weechat::buffer::{Buffer, BufferHandle, BufferLine, LineData};
 
 use crate::{render::RenderedEvent, utils::ToTag};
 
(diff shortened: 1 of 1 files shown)


strk:
Much better ! I'll merge as-is, but next I'd like to understand why the room name is NEVER updated.

Image

strk:
The issue about further improvement: https://github.com/poljar/weechat-matrix-rs/issues/176

Komzpa:
Added the follow-up as #177: https://github.com/poljar/weechat-matrix-rs/pull/177

It refreshes buffer short names when `m.room.canonical_alias` changes and makes the sync loop forward Matrix SDK `State::After` state batches too, not only `State::Before`.

PR #178, opened 2026-07-03
fix: rename Debian package

### LINKED ISSUE #175: The debian package should be called weechat-matrix-rs, not weechat-matrix
A weechat-matrix pakcage already exists in debian and is the python script.
Having the rust plugin with the same name is confusing.

@poljar do you agree or do you think the Rust plugin should REPLACE the python script ?

Komzpa:
Fixes #175.

Renames only the generated Debian package to `weechat-matrix-rs` so it does not collide with Debian's existing `weechat-matrix` package for the Python script. The Rust crate name and WeeChat plugin file remain unchanged.

Validation:
- `cargo fmt --check`
- `git diff --check upstream/main..HEAD`
- `/tmp/cargo-deb-validate/bin/cargo-deb --no-build --no-strip --deb-version 0.0.0-test` with a dummy `target/release/libmatrix.so`; `dpkg-deb -f` reported `Package: weechat-matrix-rs`

Fix requested by @strk.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,7 +93,7 @@ jobs:
       - name: Find .deb artifact
         id: deb_artifact
         run: |
-          deb_file=$(ls target/debian/weechat-matrix_*.deb)
+          deb_file=$(ls target/debian/weechat-matrix-rs_*.deb)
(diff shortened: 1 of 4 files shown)


strk:
Kisses to your minions !

PR #179, opened 2026-07-05
Change plugin install path in debian package to match weechat-core

strk:
The weechat-core debian package installs plugins under

  /usr/lib/x86_64-linux-gnu/weechat/plugins/

This commit makes weechat-matrix-rs debian package do to same


strk:
NOTE: this was purely biological work (no AI). Assisted-By:  @flashcode

PR #192, opened 2026-07-08
fix: avoid scary duplicate nick errors

### LINKED ISSUE #190: ERROR: nick already added
At startup I'm getting this error printed:
>  2026-07-08T02:13:50.295594Z ERROR matrix::room::members: Error adding nick ormai (@ormai:ormai.me) to room matrix_org.!vlrIayscTbS45jAkv1ePIrZHzy7q-_CYJ6Wf_y7PJJs, already added.

It's scary for it to be an ERROR, does it need be ? Is it hiding a programming error ?

Komzpa:
Duplicate nick inserts can happen while restoring or refreshing room members. Keep a debug trace for that case, but do not log it as an `ERROR` at startup.

Fixes #190.

### DIFF
--- a/src/room/members.rs
+++ b/src/room/members.rs
@@ -2,7 +2,7 @@ use std::rc::Rc;
 
 use dashmap::DashMap;
 use tokio::runtime::Handle;
-use tracing::{error, info};
+use tracing::{debug, info};
 
 use matrix_sdk::{
     deserialized_responses::AmbiguityChange,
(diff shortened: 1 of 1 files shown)


Komzpa:
Fixed the missing `error!` import from the first pass at `acf1a87`. Validation: `cargo fmt --check`; `git diff --check`; Ubuntu 24.04 container: `cargo fmt --check`; `cargo check --lib`.

PR #197, opened 2026-07-12
fix: render plain media as an authenticated download

### LINKED ISSUE #196: Improve rendering of uploaded media
Right now if a user attaches a media, what's shown by weechat is something like this:

```
<&strk> <image.png>[³https://matrix.osgeo.org/_matrix/media/r0/download/osgeo.org/EzMtwSkDjnxyHmSetrHGeWGJ]
<&strk> --        authenticated download: /matrix media download mxc://osgeo.org/EzMtwSkDjnxyHmSetrHGeWGJ [file]
```

Given the first URL is completely useless, unless you have setup a tool that automatically injects the matrix authentication in the request to fetch the resource, I suggest we only keep the second one, maybe with a message like this:
```
 < *> strk attached `image.png`, authenticated download: /matrix media download mxc://osgeo.org/EzMtwSkDjnxyHmSetrHGeWGJ [file]
```

@Komzpa do you think minions would enjoy working on this ?

oiahoon:
## Summary

- replace the two-line plain-media rendering with one attachment message containing the authenticated download command
- stop showing the unauthenticated HTTP media URL that fails on homeservers requiring authentication
- keep encrypted media on the existing `emxc://` rendering path

The plain-media output now reads like:

```text
attached `image.png`, authenticated download: /matrix media download mxc://example.org/media-id [file]
```

Fixes #196

## Validation

- `cargo test --locked` (22 passed)
- `cargo fmt -- --check`
- `cargo clippy --locked --all-targets` (completes with existing warnings)
- `git diff --check`

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -442,6 +442,16 @@ fn media_download_command(source: &MediaSource) -> Option<String> {
     }
 }
 
+fn media_download_message<C: HasUrlOrFile>(content: &C) -> Option<String> {
+    media_download_command(content.source()).map(|command| {
+        format!(
+            "attached `{}`, authenticated download: {}",
+            content.body(),
+            command
+        )
+    })
+}
+
 impl<C: HasUrlOrFile> Render for C {
     type RenderContext = Url;
     const TAGS: &'static [&'static str] = &["matrix_media"];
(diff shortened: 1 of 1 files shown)


Komzpa (approved):
Looks good to me. The current head `dd508c30` matches
poljar/weechat-matrix-rs#196: plain media is rendered as a single attachment
line with the authenticated download command, while encrypted media keeps the
existing `emxc://` path. The added regression test covers the new plain-media
text.

GitHub Actions is waiting for maintainer approval:
https://github.com/poljar/weechat-matrix-rs/actions/runs/29189367271

PR #198, opened 2026-07-12
fix: render plain media on one line

### LINKED ISSUE #196: Improve rendering of uploaded media

Komzpa:
Plain media rendering no longer converts the MXC URI into an unauthenticated
HTTP link before adding a second line. It now shows one line with the
attachment label and authenticated `/matrix media download` command; encrypted
media keeps its existing `emxc://` link.

Regression tests cover both rendering paths.

Fixes poljar/weechat-matrix-rs#196.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -356,26 +356,6 @@ fn mxc_to_http_download_path(
     ))
 }
 
-/// Convert a matrix content URI to HTTP(s), respecting a user's homeserver
-fn mxc_to_http(
-    mxc_url: &MxcUri,
-    homeserver: &Url,
-) -> Result<String, Box<dyn std::error::Error>> {
-    let url = url::Url::parse(mxc_url.as_str())?;
-
-    if url.scheme() != "mxc" {
-        return Err("URL missing MXC scheme".into());
-    }
-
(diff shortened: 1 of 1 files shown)


Komzpa:
Closing in favor of the earlier poljar/weechat-matrix-rs#197, which implements
the same single-line authenticated-download rendering. Thanks @oiahoon.

Issue #195, opened 2026-07-09
Improve rendering of reply context


strk:
After #194 was merged we have some visual indication about the fact a message is a reply to something else, but it's not too readable:

<img width="1312" height="112" alt="Image" src="https://github.com/user-attachments/assets/8c585566-a46c-43fb-9c1c-b9d707d89425" />

For comparison, Element/web renders those as the following:

<img width="780" height="341" alt="Image" src="https://github.com/user-attachments/assets/8571f4ca-5baf-4006-879c-534736baf5e8" />

It's interesting to note that sometimes replies get some context printed, I'm not sure how/why:

<img width="1059" height="81" alt="Image" src="https://github.com/user-attachments/assets/488bf79b-f8f8-4dbe-af01-d14c0249f375" />

I'm guessing it's a user (or client) choice whether or not to quote the message being replied, but I'm not sure that's what's happening.

In any case the "Reply to xxxx" line isn't adding much usable information. The quoted replied-to text is. Could this maybe automatically quote the replied to message (maybe partially?) if no quoting is done ?

I feel this requires some planning rather than a new PR submitted straight away. Could your minions come to chat in the matrix room @Komzpa ? :)

Komzpa:
Agree this needs planning before another PR.

The split I would use:

1. Keep the cheap visible reply indicator as a fallback, so replies are still recognizable when no context is locally available.
2. Render a compact quoted snippet when the event already contains reply fallback text, or when the replied-to event is already locally known.
3. Do not make normal message rendering block on network/backfill to fetch a missing target event. If we want that later, it should be an async cache/fill-in path rather than a synchronous render-path lookup.

Does that direction match what you want before I cut code?

strk:
> 1. visible reply indicator as a fallback

If by that you mean the current text I'm not sure it's any useful. If it was an http link to actually READ the comment it could be, but other than that, how could I possibly jump from that text to the actual message that were being replied ? 

> 2. Render a compact quoted snippet when the event already contains reply fallback text

I didn't dig into the specs to understand the difference between a threaded response and a simple response referencing another message, but from your comment I understand the message could not be readily available and thus require fetching it from some possibly even remote server ? I think rendering a compact quoted snippet is good for now, when possible

> 3. Do not make normal message rendering block on network/backfill to fetch a missing target event.

I did not understand this part.

strk:
[...]

PR #205, opened 2026-07-19
fix: quote reply fallback in context

### LINKED ISSUE #203: Compile warnings
As of commit 72ef50f3ace8919013e56bfbd18f109c8565be39 with rustc 1.93.0 I get these warnings:
```
   Compiling weechat-matrix v0.1.0 (/usr/local/src/matrix/weechat-matrix-rs)
warning: unreachable pattern
   --> src/server.rs:919:13
    |
915 |             AnyToDeviceEvent::KeyVerificationDone(e) => {
    |             ---------------------------------------- matches all the relevant values
...
919 |             AnyToDeviceEvent::KeyVerificationDone(e) => {
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no value can reach this
    |
    = note: `#[warn(unreachable_patterns)]` (part of `#[warn(unused)]`) on by default

warning: method `room` is never used
   --> src/room/members.rs:274:8
    |
 56 | impl Members {
    | ------------ method in this implementation
...
274 |     fn room(&self) -> &Room {
    |        ^^^^
    |
    = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: `weechat-matrix` (lib) generated 2 warnings
```

\cc @Komzpa

Komzpa:
## Summary

- include the sanitized Matrix rich-reply fallback text in the rendered `Reply to ...` context line, so the reply indicator carries a compact preview instead of only the sender or event id
- keep stripping the `<mx-reply>` fallback from the visible message body, so the quoted text is not duplicated below
- remove the duplicate `KeyVerificationDone` arm and unused `Members::room()` helper reported in `poljar/weechat-matrix-rs#203`
- use Debian trixie's packaged WeeChat dependencies in the Debian package job, because the external WeeChat apt repository does not currently publish a `trixie` release file

## Details

Matrix replies can already contain the replied-to body in the formatted fallback. The previous renderer sanitized that fallback away before it could be reused for the context line. This keeps a compat-sanitized copy just long enough to extract the sender and body, then runs the existing Matrix sanitization for normal display.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `WEECHAT_BUNDLED=1 cargo test --locked --lib` in `rust:1.96-bookworm`
- trixie container install of the Debian package job dependencies, including `weechat-dev`

Addresses poljar/weechat-matrix-rs#195.
Fixes poljar/weechat-matrix-rs#203.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,10 +73,6 @@ jobs:
         run: |
           apt-get update
           apt-get install -y ca-certificates curl gnupg apt-transport-https build-essential pkg-config libssl-dev libsqlite3-dev libclang-dev git
-          mkdir -p /etc/apt/keyrings
-          curl --silent https://weechat.org/dev/info/debian_repository_signing_key_asc/ | tee /etc/apt/keyrings/weechat.asc
-          echo "deb [signed-by=/etc/apt/keyrings/weechat.asc] https://weechat.org/debian trixie main" | tee /etc/apt/sources.list.d/weechat.list
-          apt-get update
           apt-get install -y weechat-curses weechat-plugins weechat-python weechat-perl weechat-dev
       - name: Install Rust toolchain
         run: |
--- a/src/render.rs
+++ b/src/render.rs
@@ -41,7 +41,7 @@ use matrix_sdk::{
 };
 use ruma_html::{
     matrix::{AnchorUri, MatrixElement},
-    Html, NodeData, NodeRef,
+    Html, NodeData, NodeRef, SanitizerConfig,
 };
 
 use weechat::{Prefix, Weechat};
@@ -82,6 +82,26 @@ impl RenderedEvent {
         event_id: &EventId,
         sender: Option<&str>,
     ) -> Self {
+        let reply_fallback = self.content.reply_fallback.take();
+        let reply_sender = sender
+            .map(ToOwned::to_owned)
+            .or_else(|| {
+                reply_fallback
+                    .as_ref()
+                    .and_then(|fallback| fallback.sender.clone())
+            })
+            .unwrap_or_else(|| event_id.as_str().to_owned());
+        let message = match reply_fallback {
+            Some(fallback) => {
+                format!(
+                    "Reply to {}: {}",
+                    reply_sender,
+                    compact_reply_preview(&fallback.body)
+                )
+            }
+            None => format!("Reply to {}", reply_sender),
+        };
+
         let mut tags = self
             .content
             .lines
@@ -90,16 +110,7 @@ impl RenderedEvent {
             .unwrap_or_default();
         tags.push("matrix_reply".to_owned());
 
-        self.content.lines.insert(
-            0,
-            RenderedLine {
-                tags,
-                message: format!(
-                    "Reply to {}",
-                    sender.unwrap_or_else(|| event_id.as_str())
-                ),
-            },
-        );
+        self.content.lines.insert(0, RenderedLine { tags, message });
 
         self
     }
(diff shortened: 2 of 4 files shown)


strk-ai-agent (commented):
test

strk-ai-agent:
Reviewed locally and on the CI run for this PR.

## Summary

Four independent concerns, all handled cleanly:

- Adds the replied-to body preview to the `Reply to …` line for text-message replies (#195).
- Keeps `<mx-reply>` stripped from the visible body so the quoted text isn't duplicated.
- Drops the duplicate `KeyVerificationDone` arm and the unused `Members::room()` helper (#203).
- Stops adding the external WeeChat apt repo to the `debian:trixie` job, since `trixie` has no Release file there; relies on the in-image `weechat-*` packages instead.

## Verification

- `cargo fmt --check` — clean
- `git diff --check` — clean
- `WEECHAT_BUNDLED=1 cargo check --lib --tests` — no warnings, no errors
- CI run `#29690859353`: `41 passed; 0 failed; 0 ignored`, including both new tests:
  - `reply_context_uses_formatted_reply_fallback_preview`
  - `reply_context_uses_reply_fallback_sender_when_local_sender_is_unknown`
- Existing tests (`reply_context_names_known_sender`, `reply_context_falls_back_to_event_id`, `formatted_body_plain_text_removes_reply_fallback`, …) still pass — no behavioural regression for messages without a fallback or for plain `body` in emote/notice replies.

## Implementation walk-through

`formatted_message_body_to_plain_text` now sanitizes twice with `ruma_html`:

```rust
let html = Html::parse(body);
html.sanitize_with(&SanitizerConfig::compat());   // sanitize but KEEP mx-reply

let mut reply_fallback = None;
for node in html.children() {
    if reply_fallback.is_none() {
        reply_fallback = render_reply_fallback(&node);
    }
}

html.sanitize();   // = compat().remove_reply_fallback()
```

Clever use of the existing API: capture the reply context before the fallback element is removed, then render the rest normally. `ReplyFallback` (private) carries `sender` + `body`, is stashed on `RenderedContent::reply_fallback` (private), and consumed in `add_reply_context`, which falls back through caller-supplied sender → fallback sender → event ID (preserving the pre-PR behaviour). Body extraction walks the blockquote and renders everything after the first `<br>` using the existing `render_html_node`, so emphasis/code/lists/blockquote prefixes still come out correctly. Sender extraction collects `<a>` text nodes before the first `<br>`, filters out empty anchors and the literal `In reply to`, and takes the last one — matching the format mandated by the Matrix rich-reply spec (and used by Element Web).

## Strengths

- **Minimal surface area.** `RenderedContent` is module-internal; all internal call sites switched to `RenderedContent::new(...)` (+ builder-style `with_reply_fallback`). `formatted_body_to_plain_text(&str) -> String` keeps its signature, so `Emote`/`Notice` renderers are untouched.
- **Backwards compatible.** Messages without `mx-reply`, and emote/notice replies, behave as before.
- **No reinvented wheels.** Uses `SanitizerConfig::compat()` instead of hand-rolling HTML walking; reuses `render_html_node` for body rendering.
- **Two warnings, two strikes.** `KeyVerificationDone` arm and `room()` helper removed exactly as requested.
- **CI fix is correct.** WeeChat apt repo has no `trixie` Release; `weechat-dev` ships in `trixie` so the workaround is fine.

## Nits (non-blocking)

1. **More tests would be nice.** Add cases for: body > 160 chars (truncation with `…`), non-ASCII/emoji body (Unicode-safe counting), no `<br>` in fallback (early-return body empty), and a plain `body` reply (no `formatted_body`).
2. **`element.to_matrix().element` is allocated per node** (and called recursively in the new helpers). Mirrors existing `render_html_node` style, so not a new sin, but a profile hotspot candidate if a very large reply is ever rendered.
3. **`render_reply_fallback_sender` heuristic.** Add a one-line doc comment pointing at the Matrix spec, so the assumption is explicit for future readers.
4. **`collapse_whitespace` only collapses spaces**, not newlines. On a single-line WeeChat preview this is fine, but worth a thought if multi-paragraph fallback bodies ever surface a stray newline in the preview.
5. **`find_first_matrix_reply` / `find_first_blockquote`** are nearly identical; could be collapsed into one generic `find_first<F: Fn(MatrixElement) -> bool>`. Minor.
6. **`ReplyFallback` and `RenderedContent::reply_fallback`** would benefit from one-line `///` doc comments.
7. **`release.yml`:** consider dropping a one-line comment in the workflow (`# weechat.org/debian has no trixie Release; use the in-image packages`) so a future contributor doesn't try to re-add the apt source.

## Verdict

**Approve.** The PR does exactly what its description says, no more and no less. The main change is well thought out, uses the upstream `ruma_html` API idiomatically, the warnings in #203 are removed as requested, and the CI is unblocked on trixie. The nits above are optional polish; happy to land as-is and follow up with extra tests in a separate commit if desired.

Issue #100, opened 2025-04-26
Make Matrix Server Come off Main Buffer / Sorting


JRGonz:
If you connect to irc servers you can detach them from the main buffer (buffer 1) and organize them into trees. weechat-matrix-rs does not seem to want to detach. I'm not sure if this is a weechat issue or a weechat-matrix-rs issue. 

Sorting doesn't seem to work properly as well but I suspect that is because of the underlying real names of the channels and it is not sorting by displayed names.

panekj:
Have you tried changing `matrix-rust.look.server_buffer` config option?

JRGonz:
So sorry. I didn't realize it had its own separate option. Yes! That worked for making it independent. I'm still having issues with sorting working properly though because it seems to use common names for channels.

PR #206, opened 2026-07-20
Add Matrix thread buffers

### LINKED ISSUE #166: Add support for threads
I'm not sure what the best UX would be for threads, I can think two ways:

  1. Each thread gets some prefix (before sender nickname?)
  2. Each thread gets its buffer named with the original room prefix

Number 2 would probably be easier to interact with (you just type in the new buffer to interact with the thread).
Number 1 would keep everything togheter but would require you to be explicit in where you want to talk.

I remember in irssi times there was a way to have multiple channels/users in the same buffer, not sure if this is still the case with weechat, if that's still a possiblity we could use solution 2 and let users possibly get the experience of 1 by merging those related buffers togheter.

\cc @Komzpa (hey, I hope you're actually giving weechat at try, it's a wonderful piece of software !)

Komzpa:
## Summary

This adds Matrix thread support using separate WeeChat buffers for each thread, matching the second UX option discussed in poljar/weechat-matrix-rs#166.

Threaded incoming messages now open or reuse a thread buffer derived from the room buffer and the thread root event id. Typing in that thread buffer sends an `m.thread` relation back to Matrix, using the latest known event in that thread as the fallback target.

The room-level buffer ownership and rendered-line lookup paths now also know about thread buffers, so local echo replacement, edit replacement, and duplicate detection still work when a message belongs to a thread.

Fixes poljar/weechat-matrix-rs#166.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `WEECHAT_BUNDLED=1 cargo test --locked --lib thread` in the Rust 1.96 Bookworm container

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -122,12 +122,8 @@ impl Servers {
             }
 
             for room in server.rooms() {
-                let buffer_handle = room.buffer_handle();
-
-                if let Ok(b) = buffer_handle.upgrade() {
-                    if buffer == &b {
-                        return BufferOwner::Room(server.clone(), room);
-                    }
+                if room.owns_buffer(buffer) {
+                    return BufferOwner::Room(server.clone(), room);
                 }
             }
 
--- a/src/room/buffer.rs
+++ b/src/room/buffer.rs
@@ -1,9 +1,12 @@
-use std::{borrow::Cow, cell::RefCell, rc::Rc};
+use std::{borrow::Cow, cell::RefCell, collections::HashMap, rc::Rc};
 
 use futures_util::StreamExt;
 use matrix_sdk::{
     room::ParentSpace,
-    ruma::{EventId, OwnedRoomAliasId, OwnedUserId, TransactionId, UserId},
+    ruma::{
+        EventId, OwnedEventId, OwnedRoomAliasId, OwnedUserId, TransactionId,
+        UserId,
+    },
     Error, Room,
 };
 use tokio::runtime::Handle;
@@ -19,6 +22,7 @@ pub struct RoomBuffer {
     room: Room,
     runtime: Handle,
     pub(super) inner: Rc<RefCell<Option<BufferHandle>>>,
+    thread_buffers: Rc<RefCell<HashMap<OwnedEventId, BufferHandle>>>,
 }
 
 impl RoomBuffer {
@@ -27,6 +31,7 @@ impl RoomBuffer {
             room,
             runtime,
             inner: Rc::new(RefCell::new(None)),
+            thread_buffers: Rc::new(RefCell::new(HashMap::new())),
         }
     }
 
@@ -44,6 +49,39 @@ impl RoomBuffer {
             .as_ref()
             .and_then(|b| b.upgrade().ok().map(|b| b.short_name().to_string()))
             .unwrap_or_default()
+    }
+
+    pub fn owns_buffer(&self, buffer: &Buffer) -> bool {
+        if self
+            .inner
+            .borrow()
+            .as_ref()
+            .and_then(|b| b.upgrade().ok())
+            .is_some_and(|b| &b == buffer)
+        {
+            return true;
+        }
+
+        self.thread_buffers
+            .borrow()
+            .values()
+            .any(|handle| handle.upgrade().is_ok_and(|b| &b == buffer))
+    }
+
+    pub fn thread_buffer(&self, thread_root: &EventId) -> Option<BufferHandle> {
+        self.thread_buffers.borrow().get(thread_root).cloned()
+    }
+
+    pub fn set_thread_buffer(
+        &self,
+        thread_root: OwnedEventId,
+        handle: BufferHandle,
+    ) {
+        self.thread_buffers.borrow_mut().insert(thread_root, handle);
+    }
+
+    pub fn remove_thread_buffer(&self, thread_root: &EventId) {
+        self.thread_buffers.borrow_mut().remove(thread_root);
     }
 
     /// Return the sender ID for an event that is still in the buffer.
@@ -64,16 +102,19 @@ impl RoomBuffer {
 
     /// Return whether an event is already rendered in the buffer.
     pub fn contains_event(&self, event_id: &EventId) -> bool {
-        let buffer_handle = self.buffer_handle();
-        let Ok(buffer) = buffer_handle.upgrade() else {
-            return false;
-        };
         let event_id_tag = Cow::from(event_id.to_tag());
 
-        buffer
-            .lines()
-            .rfind(|line| line.tags().contains(&event_id_tag))
-            .is_some()
+        let main_buffer_contains_event = self
+            .buffer_handle()
+            .upgrade()
+            .is_ok_and(|buffer| buffer_contains_tag(&buffer, &event_id_tag));
+
+        main_buffer_contains_event
+            || self.thread_buffers.borrow().values().any(|handle| {
+                handle.upgrade().is_ok_and(|buffer| {
+                    buffer_contains_tag(&buffer, &event_id_tag)
+                })
+            })
     }
 
     /// Replace the local echo of an event with a fully rendered one.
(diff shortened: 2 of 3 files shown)


strk:
I've given this branch a try, here's what I see when a new thread is created:

image

What can be improved:

# Buffer naming

The long name of that thread room makes my buffer list only 2 columns, which is degrading the UI. Maybe the buffer name could be configurable, with the default being something short like a ` (thread)` suffix, assuming weechat allows multiple buffers with the same name. The configurable suffix should allow for placeholders to indicate the thread identifier (which I think is what you are using now) or a substring of the message which originated the thread (which is what the Element/Web is using):
  
image

# Buffer placement

As you see in the screenshot the thread buffer is placed at the same indentation level as the server buffers.
Ideally it would be placed UNDER the parent buffer "#OSGeo Lounge" and further indented

# Buffer description

This is what I get as buffer description:

image

The description of the buffer could contain the full message that originated the thread, rather than (or in addition to) the identifier of the thread. Honestly I think the identifier isn't that useful for humans, so I'd keep it just available in a buffer variable.

# Tab-completion for mentioning users

This I believe would be harder to implement so not sure it's worth tackling at this stage, but it is the ability to tab-complete mentions of users. The thread belongs to a room which is full of users, mentioning a room user from within the thread would be a natural thing to do.

image

image

# Conclusions

I love this work, and I hope your minions will be able to read the screenshots too ! (must be hard for them, give them a treat from me)

strk:
Another thing: including the originating message in the thread buffer is probably a better idea rather than using it as the buffer description, see why:

image

I saw `4` but had no idea what that was about.
Compare with Element/Web:

From the outside (think "buffer list"):

image

From the inside (think "buffer content"):

image

strk-ai-agent:
## Review & Recommended Action Plan

Great work on the core threading logic -- the `m.thread` relation sending, message routing, echo/edit replacement across thread buffers, and the clean helper function extractions are all solid. The remaining work is mostly presentation-layer polish.

### Phase 1: Merge-blocking fixes

| # | Issue | Effort | Details |
|---|-------|--------|---------|
| 1 | **Buffer description** | Low | In `get_or_create_thread_buffer()`, replace `"Thread {} in {}", thread_root, room_short_name` with a description showing the originating message. At minimum, fetch the root event content and use it as the buffer title instead of the raw event ID. The event ID should only live in a buffer localvar. |
| 2 | **Buffer naming** | Medium | Change `calculate_thread_buffer_name()` to produce a shorter name. Options: (a) hardcode a short `(thread)` suffix appended to the room short name, (b) add a config option like `thread_name_format` with a default such as `{room}.thread`. For v1, at minimum shorten to `{short_room_name}.thread.{short_id}` rather than using the full sanitized event ID. |

### Phase 2: Should-fix improvements

| # | Issue | Effort | Details |
|---|-------|--------|---------|
| 3 | **Buffer placement** | Medium-High | Thread buffers currently appear at the same level as server/room buffers in the buffer list. Use a numbering scheme to place them after their parent room buffer (e.g., if `#foo` is buffer #100, threads could be #1000xx). WeeChat reorders by buffer number, so this is the primary mechanism for hierarchical display. |
| 4 | **Show originating message in thread content** | Medium | When a thread buffer is first created, fetch the root event content and print it as the first line in the buffer (matching Element/Web behavior). This gives immediate context -- without it, the user sees an opaque number like "4" with no idea what the thread is about. |

### Phase 3: Nice-to-have (future work)

| # | Issue | Effort | Details |
|---|-------|--------|---------|
| 5 | **Tab-completion for @mentions** | High | Wire up WeeChat's nick completion for thread buffers by ensuring the buffer inherits the room's member list. May need a custom completion callback. |
| 6 | **Configurable thread name format** | Low | Add a config option with placeholders like `{room}`, `{message}` (first N chars of root message), `{id}` (short event ID). |

The core threading mechanics don't need changes -- only the presentation layer needs polish before merge.

strk:
> WeeChat reorders by buffer number, so this is the primary mechanism for hierarchical display.

Did you verify this by reading the weechat source code ? I'm pretty sure I'm using a configuration (or script?) to deal with the sorting of buffers so each buffer appears indented under the buffer of the server it belongs to, do you think that configuration is changing buffer numbers ?

strk-ai-agent:
Good catch -- no, I didn't verify that. I was wrong about the mechanism.

WeeChat's default buffer ordering is by number, but the indentation you're seeing comes from the **buffer name hierarchy** using `.` as a separator. Scripts like `buffers.pl` (and the built-in `buffers` plugin) sort by name and derive indentation from the `.` delimiters in the buffer name.

So the fix for buffer placement is actually simpler than I suggested -- it falls out naturally from fixing the buffer naming (item #2 in the plan). As long as the thread buffer name starts with the parent buffer name + `.` (e.g., `#OSGeo Lounge.thread.abc123`), name-based sorting scripts will group and indent it under the parent.

No need to touch buffer numbers at all. I'll strike item #3 from the plan -- it's really a subset of item #2.

strk:
> Scripts like `buffers.pl`

According to `nils_2` in the #weechat IRC channel on libera.chat buffers.pl script is outdated and not under development, since years.

Personally I'm using the `autosort.py` script: https://weechat.org/scripts/source/autosort.py.html/

Do you know how could this work play nicely with autosort.py ?

strk-ai-agent:
Looked into `autosort.py` (v3.10). Good news -- the current naming scheme in this PR already plays nicely with it.

`autosort.py` works by evaluating a list of WeeChat expressions against each buffer, producing a sort key, then assigning buffer numbers accordingly. The default rules are (in priority order):

1. `${core_first}` -- core.weechat first
2. Script/plugin type ordering (`core, *, irc, bitlbee, matrix, slack`)
3. `${script_or_plugin}` -- group by plugin
4. `${irc_raw_first}` -- irc_raw before other irc buffers
5. `${server}` -- group by server
6. `${type}` ordering (`server, *, channel, private`)
7. `${hashless_name}` -- buffer name without `#`
8. `${buffer.full_name}` -- full internal name

For a thread buffer named `#OSGeo Lounge.thread.$abc123`:
- `${server}` = same as the parent room (both belong to the same matrix server)
- `${hashless_name}` = `OSGeo Lounge.thread.abc123` -- this sorts **alphabetically right after** `OSGeo Lounge` (the parent room), since `.` comes after nothing in string comparison
- `${buffer.full_name}` = `matrix.#OSGeo Lounge.thread.abc123` -- same, sorts after the parent

So the name-based sorting naturally places thread buffers immediately after their parent room, with no config changes needed.

The one thing to be aware of: the `${type}` rule uses `server, *, channel, private`. If the thread buffer's type is "thread", it falls into the `*` wildcard (position 1) and would sort before channels. Users who want threads after channels can add a rule like `server, *, channel, private, thread` -- but this is a user-side config tweak, not something the PR needs to worry about.

TL;DR: The current naming convention (`{room}.thread.{event_id}`) is already autosort-compatible. The `.` hierarchy is exactly what the script uses for natural grouping.

strk-ai-agent:
Found the culprit. The `.` naming convention is fine for autosort.py, but the `type` localvar is the problem.

In `get_or_create_thread_buffer`, the PR sets:
```rust
buffer.set_localvar("type", "thread");
```

autosort.py's default sorting rule for type is:
```
${info:autosort_order,${type},server,*,channel,private}
```

Since `"thread"` isn't in the list, it matches `*` (position 1), which sorts **after server but before channel**. That's why threads appear at the same indentation level as server buffers.

The fix is straightforward: either remove the `type = "thread"` localvar (matrix room buffers don't set one either, so threads would sort with them), or change it to something that sorts after channels. The simplest v1 fix is to just remove that line — the `thread_root` localvar already identifies thread buffers for the plugin's own logic.

strk:
> `"thread"` isn't in the list, it matches ``

Would it make sense to contribute an addition to `autosort.py` too, once we merge this in ? Or is it too matrix-specific to be worth the trouble ?

strk-ai-agent:
I'd skip upstreaming to autosort.py -- it's a one-line config change for users:

```
/autosort rules update 6 ${info:autosort_order,${type},server,*,channel,private,thread}
```

Adding "thread" to the default rules in autosort.py would be a bit matrix-specific (IRC doesn't have threads, and the other supported plugins like bitlbee/slack don't either). The script is general-purpose by design.

Better to document it as a recommended config tweak in the weechat-matrix-rs wiki/README alongside this feature, so matrix users know to add it.

Komzpa:
Pushed a small thread-buffer presentation update to `poljar/weechat-matrix-rs#206`.

Current head is `9dc92a8b2e58aa1d5678d66172a27610df745f39`. The new commit shortens the event-id suffix used in thread buffer names and stops setting `type=thread`, so default autosort rules no longer put thread buffers before channel buffers.

Local validation passed:

- Docker `rust:1.96-bookworm`, `WEECHAT_BUNDLED=1 cargo test --locked thread` (3 tests)
- `cargo fmt --check`
- `git diff --check`

Hosted checks are running on the new head: https://github.com/poljar/weechat-matrix-rs/pull/206/checks

PR #211, opened 2026-07-22
feat: add manual SSO token completion

### LINKED ISSUE #172: Support manual SSO token completion fallback
The smaller SSO implementation in poljar/weechat-matrix-rs#154 uses the Matrix SDK `sso-login` callback flow and intentionally does not include a manual token completion command.

A manual fallback is still useful for cases where the local callback flow cannot complete cleanly, or where users copy a `loginToken` from a browser redirect and want to finish login from WeeChat.

Possible shape:

- add a command such as `/matrix sso-complete <server> <login-token>`;
- complete Matrix login with the supplied token;
- persist the resulting session the same way as the SDK SSO flow;
- document that this is a fallback, not the primary SSO path.

This is split out of poljar/weechat-matrix-rs#154 and the older closed poljar/weechat-matrix-rs#119 discussion.

Komzpa:
Adds `/matrix sso-complete <server-name> <login-token>` for the manual SSO fallback requested in poljar/weechat-matrix-rs#172.

The command lets a user paste a `loginToken` from a browser redirect when the local callback flow cannot finish cleanly. It completes Matrix token login on the configured server, stores the returned device id through the same persistence path as normal login, and then starts the normal sync connection.

It also handles the pending-callback case from `/matrix connect`: if a connection object exists but the client is not logged in yet, the pending connection is dropped before the copied token is consumed.

Fixes poljar/weechat-matrix-rs#172.

Validation:

- `cargo fmt --check`
- `git diff --check`
- Docker `rust:1.96-bookworm` with `WEECHAT_BUNDLED=1 cargo check --locked`

### DIFF
--- a/src/commands/matrix.rs
+++ b/src/commands/matrix.rs
@@ -42,6 +42,7 @@ impl MatrixCommand {
             .add_argument("media download <mxc-uri> [file]")
             .add_argument("disconnect <server-name>")
             .add_argument("reconnect <server-name>")
+            .add_argument("sso-complete <server-name> <login-token>")
             .add_argument("read")
             .add_argument("help <matrix-command> [<matrix-subcommand>]")
             .arguments_description(format!(
@@ -50,6 +51,7 @@ impl MatrixCommand {
   disconnect: Disconnect from one or all Matrix servers.
    reconnect: Reconnect to server(s).
        join: Join a Matrix room by ID or alias.
+sso-complete: Finish SSO login with a copied loginToken.
        read: Mark the current room as read.
      devices: {}
         keys: {}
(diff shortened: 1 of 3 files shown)


Komzpa:
Rebased onto current `upstream/main` (`5e2dde414bf9133f8deb150e86662820b32af1a3`) after the latest merges. Current head is `7a92ff31776896ba1659b2db8b2f2c4717f21f30`.

The range-diff is identical to the previous PR commit. `cargo fmt --check`, `git diff --check`, and Docker `rust:1.96-bookworm` with `WEECHAT_BUNDLED=1 cargo test --locked --lib` (52 tests) pass. Ready to merge.

src/commands/matrix.rs
```diff
@@ -50,6 +51,7 @@ impl MatrixCommand {
   disconnect: Disconnect from one or all Matrix servers.
    reconnect: Reconnect to server(s).
        join: Join a Matrix room by ID or alias.
+ sso-complete: Finish SSO login with a copied loginToken.
```
strk: should all lines have the `:` character on the same columns, when printed on a terminal with fixed-width fonts ?

strk (commented):
Nitpicking the LLM. Shaving the AI yak. Will the loop/goal pick this up ?

src/commands/matrix.rs
```diff
@@ -50,6 +51,7 @@ impl MatrixCommand {
   disconnect: Disconnect from one or all Matrix servers.
    reconnect: Reconnect to server(s).
        join: Join a Matrix room by ID or alias.
+sso-complete: Finish SSO login with a copied loginToken.
```
strk, line 54: it's still not aligned ?

strk (commented):
more yak shaving (did you even read this ?)

PR #215, opened 2026-07-22
fix: suppress duplicate restored backlog events

### LINKED ISSUE #185: Duplicated lines in backlog
This sounds like an off-by-one bug with the backlog manager ?

Image

Komzpa:
Restored rooms can receive sync events while member state is being loaded. The saved history cursor was installed only after those awaits, so it could overwrite a newer cursor established by sync and make backlog pagination print the same events again.

This installs the saved cursor before the first await. It also skips ordinary sync and backlog events whose Matrix event ID is already present in the room or thread buffers, while leaving redactions, verification events, and edits on their existing handling paths.

The restore-ordering insight comes from Jakub Panek's draft poljar/weechat-matrix-rs#114.

Fixes poljar/weechat-matrix-rs#185.

Checked with `cargo fmt --check`, `git diff --check`, and Docker `rust:1.96-bookworm` with `WEECHAT_BUNDLED=1 cargo test --locked --lib`.

Fix requested by @strk.

### DIFF
--- a/src/room/mod.rs
+++ b/src/room/mod.rs
@@ -116,6 +116,10 @@ fn restored_prev_batch(prev_batch: Option<String>) -> Option<PrevBatch> {
     prev_batch.map(PrevBatch::Backwards)
 }
 
+fn should_render_event(already_rendered: bool) -> bool {
+    !already_rendered
+}
+
 impl Deref for RoomHandle {
     type Target = MatrixRoom;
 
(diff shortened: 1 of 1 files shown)


Komzpa:
@strk exact head `c6d768022530fa280b4200517103661f70bbfa70` is clean and mergeable. All four hosted `build-test-release` jobs succeeded, and the Docker library suite passed 53/53. Ready to merge.

PR #216, opened 2026-07-22
feat: select /redact targets by message pattern

### LINKED ISSUE #213: Accept a pattern as the first parameter of /redact
Followup of #208: accept a pattern as first arg in `/redact`: most recent message matching the pattern would be redacted.

How to distinguish pattern is left to decide to whoever implements this.

\cc @Komzpa

Komzpa:
Adds pattern selection to `/redact` without changing the existing event-ID, index, or plain-reason forms.

`/redact /pattern/ [reason...]` scans the current buffer from newest to oldest and selects the most recent unredacted message containing the case-sensitive literal substring. WeeChat color codes are removed before matching. An empty `//` pattern is rejected.

Fixes poljar/weechat-matrix-rs#213.

Checked with `cargo fmt --check`, `git diff --check`, and Docker `rust:1.96-bookworm` with `WEECHAT_BUNDLED=1 cargo test --locked --lib`.

Fix requested by @strk.

### DIFF
--- a/src/commands/redact.rs
+++ b/src/commands/redact.rs
@@ -11,6 +11,10 @@ use crate::Servers;
 
 const EVENT_ID_TAG_PREFIX: &str = "matrix_id_";
 
+fn message_matches_pattern(message: &str, pattern: &str) -> bool {
+    message.contains(pattern)
+}
+
 pub struct RedactCommand {
     servers: Servers,
 }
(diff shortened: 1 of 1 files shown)


Komzpa:
@strk exact head `04fe8d23556079758aef1776e85ce1bdde9decb1` is clean and mergeable. All four hosted `build-test-release` jobs succeeded, and the Docker library suite passed 55/55. Ready to merge.

PR #217, opened 2026-07-22
fix: reject /ignore in Matrix buffers

### LINKED ISSUE #214: Add support for /ignore
At the moment if you type something like `/ignore add llm-gemma` in a matrix room you get a system message like:

```
irc: ignore added:
   [1] mask: ^llm-gemma$ / server: * / channel: *
```

This makes no sense, because we were not in an IRC buffer.
This ticket is to support a matrix-specific `/ignore` command, if that makes any sense.
Or as an alternative print a message (in the same buffer from which the command is given) about `/ignore` being unsupported in matrix buffers.

I don't know if the python version supported this ( @poljar ? )

\cc @Komzpa

Komzpa:
Running `/ignore` from a Matrix buffer currently falls through to the IRC
plugin and creates a global IRC ignore. Intercept both the bare command and
argument forms on Matrix-owned buffers, print an explicit unsupported message
in the same buffer, and preserve native `/ignore` behavior elsewhere.

Fixes https://github.com/poljar/weechat-matrix-rs/issues/214.

Validation:

- `WEECHAT_BUNDLED=1 cargo test --locked --lib`
- WeeChat 4.4.4 FIFO smoke covering Matrix rejection, an unchanged empty IRC
  ignore list, and a successful native IRC ignore

### DIFF
--- a/src/commands/ignore.rs
+++ b/src/commands/ignore.rs
@@ -1,0 +1,74 @@
+use std::borrow::Cow;
+
+use weechat::{
+    buffer::Buffer,
+    hooks::{CommandRun, CommandRunCallback},
+    Prefix, ReturnCode, Weechat,
+};
+
+use crate::PLUGIN_NAME;
+
+pub struct IgnoreCommand;
+
+#[derive(Debug, Eq, PartialEq)]
+enum IgnoreDisposition {
+    EatWithUnsupportedMessage,
+    PassThrough,
+}
+
+impl IgnoreCommand {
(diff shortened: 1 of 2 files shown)


Komzpa:
@strk this is ready to merge at `b63ef5bc59dd68626be9309bdf19dbc8346f7ac2`.
Both bare `/ignore` and `/ignore ...` are rejected in Matrix buffers, and the
native command still works outside Matrix. All four hosted jobs pass on this
exact head, the WeeChat 4.4.4 FIFO smoke passes, and there are no unresolved
review threads.

PR #226, opened 2026-07-26
feat: improve thread buffer UX

### LINKED ISSUE #219: Improve thread support UX
Things that need to be done to improve the UX of Matrix threads support follow. For @Komzpa 

# Include the triggering question as the first message in the buffer

Lack of that triggering question makes it impossible to tell where the conversation started, see pictures for detail

Image

# Make default buffer's short_name play nice with autosort.py

I'd like the thread buffer to be sorted and indented below its respective room.

See https://github.com/weechat/scripts/blob/main/python/autosort.py

# Make buffer short_name as short as possible

Otherwise the buffer list is going to overflow the available space as each column containing a long short_name buffer will be as large as that (supposedly short) name.

Komzpa:
## Summary

- copy the tagged thread root event from the room buffer into a newly-created thread buffer when it is still available locally
- shorten thread buffer short names from `<room>.thread.<suffix>` to `<room>.<suffix>` so autosort.py keeps them grouped under the room with less buflist width
- keep the internal full buffer name unchanged

## Validation

- `cargo fmt --check`
- `git diff --check`
- `WEECHAT_BUNDLED=1 cargo test --locked --lib` in `rust:1.96-bookworm` with `clang`, `libclang-dev`, `libsqlite3-dev`, and `pkg-config` installed: 59 passed

Closes poljar/weechat-matrix-rs#219

### DIFF
--- a/src/room/buffer.rs
+++ b/src/room/buffer.rs
@@ -119,6 +119,19 @@ impl RoomBuffer {
             })
     }
 
+    /// Copy the root event line into a newly created thread buffer.
+    pub fn seed_thread_buffer(
+        &self,
+        thread_root: &EventId,
+        thread_buffer: &Buffer,
+    ) -> bool {
+        let event_id_tag = Cow::from(thread_root.to_tag());
+
+        self.buffer_handle().upgrade().is_ok_and(|buffer| {
+            copy_buffer_line_by_tag(&buffer, thread_buffer, &event_id_tag)
+        })
+    }
+
     /// Replace the local echo of an event with a fully rendered one.
     pub fn replace_local_echo(
         &self,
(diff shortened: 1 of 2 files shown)


strk:
The first message in buffer works great, unfortunately the shorter buffer name didn't help with my autosort, anyway it makes the buffer name smaller so so it's an improvement.

While using I found another issue being that the go.py script would find threads before the parent room when typing a substring of the parent room name like `/go lounge` so now I'm thinking if it isn't better to put the parent room name last and use a "thread" prefix, like: " thread 1 #OSGeo Lounge". But before changing the short name again I'll need to figure out how to make autosort.py work

Issue #112, opened 2025-09-16
[Feature Request] add support for replying


killertofus:

SuicideShow:
And also make it visual when messages are normal messages or replies, like have a quote of the message that was replied to.
At the moment there is no difference visible, every message looks the same.

strk:
How was this completed @killertofus ? 

I still don't see any visual indication here:

<img width="1254" height="63" alt="Image" src="https://github.com/user-attachments/assets/78da0820-f120-437e-85de-437807282945" />

When \Element/Web renders it instead like this:

<img width="1136" height="138" alt="Image" src="https://github.com/user-attachments/assets/a05ba071-7d58-497b-9874-82a5193f6f74" />

@Komzpa do you have anything for me here ?

Komzpa:
#194 is merged now and adds a first visible reply indicator: replies get a `Reply to <event_id>` line before the rendered message.

That intentionally stops short of rendering the quoted event body like Element does. If that richer quote rendering is wanted, it should be a separate follow-up from the basic visible-indicator fix.

strk:
I'm reopening this as `/reply` comand sent to a matrix buffer results in the message `/reply` being sent to the room itself !!

Komzpa:
The basic reply command and current thread/reply-context work are carried by merged #268 and open PR #270 at head 92b18c2. The branch is clean; merge/review is still pending.

strk:
@ghost are you happy with the outcome of this ? Re-open if you're not

PR #227, opened 2026-07-26
feat: render recent Matrix replies inline

### LINKED ISSUE #222: Improve Reply UX
At the moment replies are rendered using 2 messages:
    - a `Reply to $user'
    - the actual reply

See screenshot:

Image

The typical IRC reply is instead all on a single line, example:

> Regina Obe: Eh, that was about not running the migration tests on every PR

In absence of more information, the one-liner would be my preference.

In some cases you DO want more information about the reply, especially if the message being replied to was in the far past so more likely to have scrolled out of your screen. In those cases it could be nice to include the replied-to message, like is done by Element/Web:

Image

another possible option would be to add some visual sign (color?icon?) to associate a reply to the message replied-to, but when the replied-to message has scrolled out of the visible screen it won't really help so I tend to prefer the idea of including the replied-to message (optionally in reduced form, if it's too long).

So my proposal:

 1. Add a variable to specify how "old" a replied-message should be before (how many messages have been seen in the buffer after that) for it to be fully included (defaulting to some sensible value possibly computed from number of LINES in the terminal ?)
 2. Use single-line when the replied-to message is more recent than the value specified with the variable

\cc @Komzpa

Komzpa:
## Summary

- keep Matrix reply fallback quotes by default by setting `matrix-rust.look.reply_full_quote_threshold` to `0`
- render recent Matrix replies on one compact line as `> Alice: message` only when the threshold is set to a positive value and the replied-to event is still within that buffer distance
- keep full quoted fallback rendering for older replies or replies whose target is no longer available in local buffers
- document the reply threshold option

## Validation

- `git diff --check`
- `WEECHAT_BUNDLED=1 CARGO_TARGET_DIR=/target /usr/local/cargo/bin/cargo test --locked --lib` in `rust:1.96-bookworm` with `clang`, `libclang-dev`, `libsqlite3-dev`, `pkg-config`, and `weechat-dev` installed: 63 passed
- hosted `build-test-release` jobs succeeded on head `6400929df6b58c533c8495d9f17a32db531dfc6a`

Closes poljar/weechat-matrix-rs#222.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -136,6 +136,17 @@ config!(
             // Range: 0 ms to one day.
             0..86400000,
         },
+
+        reply_full_quote_threshold: Integer {
+            // Description
+            "Render replies to recently printed events on one line when this \
+             value is positive. The default 0 keeps the Matrix reply fallback \
+             quote when available",
+            // Default value.
+            0,
+            // Range: quote-preserving default to a large scrollback window.
+            0..100000,
+        },
     },
 
     Section color {
--- a/README.md
+++ b/README.md
@@ -195,6 +195,13 @@ The inactivity delay defaults to five minutes and can be changed with:
 
        /set matrix-rust.look.smart_filter_delay 300000
 
+Recent replies can be rendered on one line in the form `> Alice: message`. By
+default, weechat-matrix-rs keeps the Matrix reply fallback quote when one is
+available. Set a positive threshold to render replies to recently printed lines
+in the compact one-line form:
+
+       /set matrix-rust.look.reply_full_quote_threshold 20
+
 # Room Status And Verification
 
 The `buffer_modes` bar item shows Matrix room state. Encrypted rooms show
--- a/src/render.rs
+++ b/src/render.rs
@@ -56,6 +56,12 @@ pub struct RenderedEvent {
     pub content: RenderedContent,
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum ReplyContext {
+    Inline,
+    Full,
+}
+
 impl RenderedEvent {
     const MSG_TAGS: &'static [&'static str] = &["notify_message"];
     const SELF_TAGS: &'static [&'static str] =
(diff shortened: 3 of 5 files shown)


strk:
How does this fix #112 ? I think that one is about using `/reply` ? Or what does this PR contain that you think would fix #112 ?

src/config.rs
```diff
@@ -136,6 +136,17 @@ config!(
...
             0..86400000,
         },
+
+        reply_full_quote_threshold: Integer {
+            // Description
+            "How many printed lines after a replied-to event still count as a \
+             recent reply that should be rendered on one line. Older replies \
+             include the Matrix reply fallback quote when available",
+            // Default value.
+            20,
```
strk: Please change the default to 0 (always include the quoted reply)

strk (changes requested):
Change the default, and please respond at the question about closing the unrelated issue

Komzpa:
You're right, this should not close poljar/weechat-matrix-rs#112. That issue also covers composing replies with `/reply`, while this PR only addresses the rendering UX requested in poljar/weechat-matrix-rs#222.

I removed the `Closes poljar/weechat-matrix-rs#112` line from the PR body. I also changed the default threshold to `0`, so the Matrix quoted reply fallback remains the default; positive values opt into compact one-line rendering for recent replied-to events.

strk (commented):
I've resolved the conflict via web UI

strk (approved):
Approving, let's wait for CI to check

strk:
There are conflicts to be resolved here

strk (approved):
I dunno how to unblock the minions (or I've heard someone call them "golems"), but I'm approving in the hope they'll resolve conflicts at some point

strk:
I've once again resolved the conflict, but it's worth a review

strk-ai-agent:
I still don't see the inlined reply text with the default value of 0 for `reply_full_quote_threshold`. Since 0 means "never show full quote" (inline everything) intuitively, but the current implementation treats 0 as "always show full quote" — is the semantics inverted?

strk:
I'll merge, double check and AFTER that I'll file a new issue if it still doesn't work

PR #255, opened 2026-07-29
Keep Matrix thread roots first and in sync

Komzpa:
Late Matrix thread roots could still be appended after the reply that opened the
thread buffer. A root sent by this client could also stay absent after its local
echo was acknowledged, and later edits updated only one rendered copy.

This keeps every rendered root line first in its thread buffer, including equal
timestamp ties, seeds acknowledged local echoes into already-open threads, and
propagates edits to both the room and thread copies.

Validation: Rust 1.96 with bundled WeeChat, `cargo test --locked --lib` (74
passed); `git diff --check`; clean Codex review of the exact one-commit diff.

Follow-up to poljar/weechat-matrix-rs#248. Addresses
poljar/weechat-matrix-rs#219.

Fix requested by @strk.

### DIFF
--- a/src/room/buffer.rs
+++ b/src/room/buffer.rs
@@ -27,6 +27,26 @@ pub struct RoomBuffer {
     thread_buffers: Rc<RefCell<HashMap<OwnedEventId, BufferHandle>>>,
 }
 
+struct LineCopy {
+    date: isize,
+    date_printed: isize,
+    tags: Vec<String>,
+    prefix: String,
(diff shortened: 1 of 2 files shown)


strk:
why is this a WIP ? I'm curious, what was the rationale to create the PR while still unfinished ?

Komzpa:
No good reason — that was my mistake. The branch was already complete and the checks are green. I have marked it ready for review.

PR #257, opened 2026-07-29
Make Matrix space buffers interactive

### LINKED ISSUE #254: Improve space buffer usefulness
At the moment space buffers are pretty useless:
 - They show nothing
 - You can't type anything

Image

It would be nice for it to show a list of rooms being part of the space, and letting you search in it and select one to join the room.
I'm thinking of an experience like the /fset one

Komzpa:
Space buffers currently open empty and cannot be used to navigate their rooms.
This turns the main Space buffer into an interactive hierarchy selector similar
to `/fset`.

- Show active `m.space.child` rooms and subspaces, honoring explicit Matrix
  order values before a stable fallback order.
- Filter the list live using the buffer input, with stable numeric selections.
- Open joined children and join unjoined children using the advertised `via`
  servers.
- Refresh the selector when Space membership changes and disable chat-only
  nicklist, multiline, and logging behavior.

Validation: Rust 1.96 with bundled WeeChat, `cargo test --locked --lib`
(`74 passed`); `git diff --check`; clean review of the exact one-commit diff.

Closes poljar/weechat-matrix-rs#254.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,7 +190,7 @@ impl SignalCallback for Servers {
                         })
                         .detach();
                     }
-                    "input_text_changed" => room.update_typing_notice(),
+                    "input_text_changed" => room.update_input(&buffer),
                     _ => (),
                 }
             }
--- a/src/room/buffer.rs
+++ b/src/room/buffer.rs
@@ -2,12 +2,14 @@ use std::{borrow::Cow, cell::RefCell, collections::HashMap, rc::Rc};
 
 use futures_util::StreamExt;
 use matrix_sdk::{
+    deserialized_responses::SyncOrStrippedState,
     room::ParentSpace,
     ruma::{
-        EventId, OwnedEventId, OwnedRoomAliasId, OwnedUserId, TransactionId,
-        UserId,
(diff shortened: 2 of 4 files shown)


strk:
Conflicts, maybe avoid multiple words in the same line, line break after each comma so there are less conflicts next time ?

Komzpa:
Updated the branch on top of current `main`; GitHub now reports it as mergeable again.

I also let `cargo fmt` split the affected import list, so that part should be less conflict-prone next time.

strk:
unfortunately it conflicts again, sorry

Komzpa:
The earlier conflict note is stale: GitHub currently reports this branch clean on head b05a164. The branch already includes the current-main merge and cargo fmt-compatible import layout; hosted checks were green at the last recorded run.

Komzpa:
Rebased from b05a164 onto current upstream/main d156ebd; conflict resolution preserves space-buffer and current thread/reply changes. New head: 87b1b9f. Clean Debian trixie (Rust 1.97.1) cargo test --lib: 111/111 passed. cargo fmt check remains baseline-blocked by pre-existing build.rs formatting on upstream/main.

PR #265, opened 2026-07-29
Expose Matrix room avatars, member profiles, and semantic mentions

Komzpa:
## Summary

- expose each Matrix room's avatar as `matrix_avatar_mxc` for relay UIs; when a one-to-one direct chat has no room avatar, use the sole other joined member's avatar
- keep rooms without an image explicit and never choose an arbitrary face for multi-user direct chats
- expose joined-room members as structured mention candidates for relay clients
- add `/matrix-send`, accepting a URL-safe base64 JSON payload and sending standard `m.mentions` metadata
- validate mentioned Matrix user IDs and cover the command payload and event content
- expose a deterministic, bounded `matrix_members_v1` profile list for native relay UIs, including exact Matrix ID, display name, disambiguated room nick, membership, role, power level, and avatar MXC URI
- keep credentials and media bytes out of relay metadata; consumers continue to download MXC content through the authenticated Matrix media command

Profile metadata is capped at 512 joined members, sorted by display name and Matrix ID, and refreshed with the existing member-localvar update path. Room-avatar metadata is initialized with the buffer and refreshed on room-avatar and membership changes.

The native sidebar consumer and aggregate UI smoke live in rnocx/WeeChatRS#18.

## Testing

- `cargo test --locked --release` with Rust 1.96 in Debian Bookworm and the WeeChat 4.9.5 API header: 98 passed
- `profile_export_is_sorted_and_bounded` covers deterministic order and the 512-member cap
- focused room-avatar tests cover explicit room images, one-to-one member fallback, and the multi-user-direct negative control

## Update: room upgrade relationships

- Expose Matrix predecessor and replacement room IDs as structured buffer local variables alongside the existing avatar, profile, and semantic-mention metadata.
- This lets relay GUIs represent an upgrade chain as one current chat with predecessor history instead of showing both physical rooms as current.
- `git diff --check` passes. The integrated backend composite containing this change passed all 101 tests, and all four exact-head build-test-release jobs are green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4698,6 +4698,7 @@ dependencies = [
 name = "weechat-matrix"
 version = "0.1.0"
 dependencies = [
+ "base64",
  "chrono",
  "clap",
  "dashmap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ default = []
 
 [dependencies]
 clap = "2.34.0"
+base64 = "0.22"
 chrono = "0.4.22"
 dashmap = "6.1.0"
 url = "2.3.1"
--- a/src/commands/mention_send.rs
+++ b/src/commands/mention_send.rs
@@ -1,0 +1,135 @@
+use std::{borrow::Cow, convert::TryFrom};
+
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
+use matrix_sdk::ruma::OwnedUserId;
+use weechat::{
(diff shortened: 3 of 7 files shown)


strk:
conflicts to be resolved

strk:
branch has conflicts that must be resolved - may I recommend https://github.com/ubergeek42/weechat-android/pulls for when you are on the go ? Just in case there's some improvement on nick tab-completion there too ;)

What are mentions metadata, btw ?

PR #268, opened 2026-07-30
Fix reply commands leaking into message bodies

Komzpa:
## Summary

- let Clap consume WeeChat's command name at `argv[0]`
- keep `/reply` and the selected event ID out of the Matrix message body
- accept the documented negative reply indexes such as `-2`

## Root cause

The reply parser enabled Clap's `NoBinaryName` mode even though WeeChat command
callbacks include `reply` as the first argument. Clap therefore exposed the
command name as message text, and `/reply $event_id message` could be sent with
that entire string as its body.

This is a follow-up to the reply support requested in
poljar/weechat-matrix-rs#112.

## Validation

- `cargo test --locked --release --lib commands::reply::tests`
- 4 reply tests passed, including command-name removal and negative indexes

### DIFF
--- a/src/commands/reply.rs
+++ b/src/commands/reply.rs
@@ -126,6 +126,16 @@ impl ReplyCommand {
         }
     }
 
+    fn parser() -> Argparse<'static, 'static> {
+        Argparse::new("reply")
+            .settings(&[
+                ArgParseSettings::AllowLeadingHyphen,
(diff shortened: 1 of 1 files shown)


strk:
Is this intentionally a draft ?

Komzpa:
Not intentionally; I marked it ready now.

The branch is clean and the current checks are green.

PR #270, opened 2026-07-30
Load complete Matrix threads and expose reply context

Komzpa:
## Summary

- keep historical `m.thread` replies out of the main room timeline
- load the complete relation history when a thread buffer opens
- paginate backwards with bounded retries, duplicate-event suppression, and repeated-token protection
- keep the root first and the replies chronological in the thread buffer
- preserve `m.thread` when `/reply` is used from a thread buffer
- add `/matrix thread <event-id>` for clients that need to open a specific thread
- resolve the referenced Matrix event when a reply is older than WeeChat's local buffer, so reply context has the real sender and quoted body
- run remote reply-context lookups on the Matrix connection runtime instead of polling Matrix SDK network futures on WeeChat's main-thread executor
- expose reply header/quote roles, sender, and target event as structured WeeChat tags while retaining readable text for ordinary WeeChat clients
- restore room history from the newest visible event after a plugin reload, then preload older pages until the buffer has a useful history window

## Why

Thread buffers previously contained only replies seen during the current sync. Older replies loaded through room history were treated as ordinary room events, so they could appear in the main room while the corresponding thread buffer stayed empty or incomplete.

There was a second routing gap: `/reply` always created a plain `m.in_reply_to` relation. When a relay client invoked it from a thread buffer, the reply was therefore posted to the main room instead of the thread.

Reply presentation had a separate history-bound failure. The plugin looked up the referenced sender only in the current WeeChat buffer and fell back to a raw `$event_id` when that line was no longer present. It now loads the referenced event through the Matrix SDK cache or homeserver when needed, then emits optional `matrix_reply_header`, `matrix_reply_quote`, `matrix_reply_sender_hex_*`, and `matrix_reply_id_*` tags. The existing human-readable `Reply to ...` and `> ...` lines remain available to clients that ignore those tags.

The Matrix SDK lookup may perform network I/O and therefore requires a Tokio reactor. Polling it directly from rust-weechat's main-thread executor produced `there is no reactor running` panics while restored reply history was rendered. The lookup now runs through the room's Matrix connection runtime and returns its result to the WeeChat-thread renderer.

Restored rooms also started backward pagination at the SDK's `last_prev_batch` token. That token points before the stored sync timeline, but the SDK does not replay that timeline into a newly restored WeeChat buffer. The result was an old or nearly empty buffer after reload. Restored rooms now request backward history from the current room end first and use the returned `end` token for older pages. Because homeservers may return much less than the requested limit, restoration continues for up to ten bounded pages until the buffer contains at least 100 lines, stopping early at the end or when a request makes no progress.

The thread buffer requests the Matrix relations endpoint for its root and routes both relation history and ordinary room-history events according to their `m.thread` relation. Explicit replies inherit the current thread root while retaining the selected event as their fallback reply target.

This builds on the thread buffers introduced by [poljar/weechat-matrix-rs#206](https://github.com/poljar/weechat-matrix-rs/pull/206) and preserves the root ordering fixes from [poljar/weechat-matrix-rs#255](https://github.com/poljar/weechat-matrix-rs/pull/255). The structured tags are consumed by the optional native presentation in [rnocx/WeeChatRS#15](https://github.com/rnocx/WeeChatRS/pull/15).

## Validation

- `git diff --check`
- Rust 1.96.1 Bookworm container against WeeChat plugin API `20260530-01`: `cargo test --locked --lib` (97 passed)
- focused tests cover newest-first bounded restored history, remote sender/body extraction, fetched-body fallback, structured reply tags, thread pagination, and thread-preserving `/reply`
- manual relay smoke test with a nine-event Matrix thread: one root and eight replies appeared once, in chronological order, while the main room retained only the root

### DIFF
--- a/src/commands/matrix.rs
+++ b/src/commands/matrix.rs
@@ -4,6 +4,7 @@ use clap::{
 };
 use url::Url;
 
+use matrix_sdk::ruma::EventId;
 use weechat::{
     buffer::Buffer,
     hooks::{Command, CommandCallback, CommandSettings},
@@ -48,6 +49,7 @@ impl MatrixCommand {
             .add_argument("sso-complete <server-name> <login-token>")
             .add_argument("history")
             .add_argument("read")
+            .add_argument("thread <event-id>")
             .add_argument("version")
             .add_argument("help <matrix-command> [<matrix-subcommand>]")
             .arguments_description(format!(
@@ -59,6 +61,7 @@ impl MatrixCommand {
 sso-complete: Finish SSO login with a copied loginToken.
      history: Load an older page of messages in the current room.
         read: Mark the current room as read.
+      thread: Open a thread buffer and load its complete history.
      version: Show version information about weechat-matrix.
      devices: {}
         keys: {}
@@ -86,8 +89,9 @@ Use /matrix [command] help to find out more.\n",
             .add_completion("disconnect %(matrix_servers)")
             .add_completion("reconnect %(matrix_servers)")
             .add_completion("sso-complete %(matrix_servers)")
+            .add_completion("thread")
             .add_completion(
-                "help server|connect|disconnect|reconnect|join|sso-complete|read|version|keys|devices|media|verify|verification",
+                "help server|connect|disconnect|reconnect|join|sso-complete|read|thread|version|keys|devices|media|verify|verification",
             );
 
         Command::new(
@@ -276,6 +280,42 @@ Use /matrix [command] help to find out more.\n",
         }
     }
 
+    fn thread_command(&self, buffer: &Buffer, args: &ArgMatches) {
+        let Some(room) = self.servers.find_room(buffer) else {
+            Weechat::print(&format!(
+                "{}{}: Run /matrix thread from a Matrix room buffer.",
+                Weechat::prefix(Prefix::Error),
+                PLUGIN_NAME,
+            ));
+            return;
+        };
+        let event_id = args
+            .value_of("event-id")
+            .expect("Event ID not set but was required");
+        let Ok(event_id) = EventId::parse(event_id) else {
+            Weechat::print(&format!(
+                "{}{}: Invalid Matrix event ID: {}",
+                Weechat::prefix(Prefix::Error),
+                PLUGIN_NAME,
+                event_id,
+            ));
+            return;
+        };
+
+        if let Some(handle) = room.open_thread_buffer(&event_id) {
+            if let Ok(thread_buffer) = handle.upgrade() {
+                thread_buffer.switch_to();
+            }
+        } else {
+            Weechat::print(&format!(
+                "{}{}: Event {} is not loaded in this room.",
+                Weechat::prefix(Prefix::Error),
+                PLUGIN_NAME,
+                event_id,
+            ));
+        }
+    }
+
     fn run(&self, buffer: &Buffer, args: &ArgMatches) {
         match args.subcommand() {
             ("connect", Some(subargs)) => self.connect_command(subargs),
@@ -328,6 +368,7 @@ Use /matrix [command] help to find out more.\n",
                 })
                 .detach();
             }
+            ("thread", Some(subargs)) => self.thread_command(buffer, subargs),
             ("version", _) => {
                 Weechat::print(&format!(
                     "{}: weechat-matrix version {} ({})",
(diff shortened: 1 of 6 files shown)


strk:
conflicts, and why is this a draft ?

Komzpa:
Merged current upstream/main (d156ebd) into this branch to clear the conflict. New head is 92b18c2; cargo fmt passes. Local cargo test is unavailable here because the host exposes unsupported OpenSSL 4.0.1; hosted CI is the validation gate.

Komzpa:
The conflict is resolved on head 92b18c2; marking this ready for review now. The branch keeps the complete-thread-history and reply-context changes; cargo fmt passes. Local cargo test remains unavailable on this host because OpenSSL 4.0.1 is unsupported by openssl-sys.

Komzpa:
Fresh validation after conflict repair: in a clean Debian trixie container, `cargo test --lib` passed (113 passed, 0 failed) at head `92b18c2a97aa68959472c448a5d90c3978398227`.

strk:
conflicts again, sorry

Komzpa:
Conflict repaired against current upstream/main (`8fd3f401`) and pushed as head `2eba0755b1228bda6a04763bf012a610abe02cdf`. Fresh validation: `cargo test --lib` passed in a clean rust:1.97-bookworm container (133 passed, 0 failed). GitHub checks are queued.

strk:
GitHub still reports it as conflicting 
screenshot06

Komzpa:
Rechecked from the GitHub API now: this PR is `MERGEABLE` / `CLEAN` at head `2eba0755b1228bda6a04763bf012a610abe02cdf`, and all four `build-test-release` jobs are successful. I do not see an unresolved review thread here, so this looks like stale conflict state in the UI rather than a current branch conflict.

strk:
But it still isn't. Shall we move the code to a non-bogus platform ?

image

Komzpa:
Rechecked again from two independent paths just now:

- GitHub API: `MERGEABLE` / `CLEAN` at head `2eba0755b1228bda6a04763bf012a610abe02cdf`
- local `git merge-tree --write-tree upstream/main refs/remotes/upstream-pr/270` against current `upstream/main` `98619dbfc6d7dcc4cc57654330f2a974eee85c2d` produced clean tree `7bfab11bd9ca6c4f570559a6fc37780587605c8f`
- all four `build-test-release` jobs are successful, and GraphQL shows 0 unresolved review threads

So I still cannot reproduce the conflict state outside that browser view. Maybe GitHub's PR page cache is stuck for your session?

strk:
> Maybe GitHub's PR page cache is stuck for your session?

It still happens with an anonymous window (Firefox).
Also from a different device.

Can YOU see that message or is it only visible to people with write access @Komzpa ?

@poljar do you see this PR as mergeable ?

jspricke:
I think the difference is that @Komzpa is checking for a merge (possible) whereas @strk is checking for a rebase (not possible). The problem is that the main branch was merged into this branch, so I propose that @Komzpa does a rebase.

Komzpa:
Rebased this branch onto current upstream/main (`98619db`) without merge commits, as suggested. New head is `12588cdeb2fd6d73b62e257b19732a74d83a5d43`; previous head was `2eba0755b1228bda6a04763bf012a610abe02cdf`.

Validation: `cargo test --lib` passed in a clean Debian trixie container with rustup stable `rustc 1.98.0`, WeeChat headers, OpenSSL dev headers, and sqlite dev headers: 136 passed, 0 failed.

GitHub is still recalculating mergeability/checks on the new head right now.

jspricke:
@Komzpa could you rebase this as well?

Komzpa:
Rechecked this one too: it is already rebased on current upstream/main `98619dbfc6d7dcc4cc57654330f2a974eee85c2d` at head `12588cdeb2fd6d73b62e257b19732a74d83a5d43`.

Local readback: `git merge-base HEAD upstream/main` returns `98619dbfc6d7dcc4cc57654330f2a974eee85c2d`, `git log --merges upstream/main..HEAD` is empty, and `git merge-tree --write-tree upstream/main HEAD` produced clean tree `42b308b667074c61cdadb15c0a9adaa2df6907ec`.

GitHub also reports `MERGEABLE` / `CLEAN`, with all four hosted package jobs successful.

strk:
Bingo! Thanks @jspricke

PR #271, opened 2026-07-30
Expose Matrix media metadata to relay clients

Komzpa:
## Summary

- attach structured media kind, filename, and MXC URI tags to rendered Matrix media
- encode tag payloads with unpadded URL-safe base64 so relay clients can decode them without delimiter ambiguity
- preserve the existing human-readable authenticated download command
- accept a quoted download destination without writing the quote characters into the filename

## Why

Native relay clients currently receive only the rendered media text. They cannot reliably distinguish an image attachment from an ordinary message or recover the authenticated MXC source needed for an inline preview.

Structured tags let clients render plain Matrix media without scraping presentation text. Normalizing the optional destination also makes relay-issued download commands safe for cache paths containing spaces.

This complements the authenticated media download command from poljar/weechat-matrix-rs#156.

## Validation

- `git diff --check`
- Rust 1.96 Bookworm container with bundled WeeChat: `cargo test --locked --lib` (93 passed)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ crate-type = ["cdylib"]
 default = []
 
 [dependencies]
+base64 = "0.22"
 clap = "2.34.0"
 base64 = "0.22"
 chrono = "0.4.22"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,6 @@ crate-type = ["cdylib"]
 default = []
 
 [dependencies]
-base64 = "0.22"
 clap = "2.34.0"
 base64 = "0.22"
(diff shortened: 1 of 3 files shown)


strk:
why is this a draft ?

Komzpa:
Marked ready for review as requested; the branch is clean and its four hosted package checks were green at the recorded head 01a8d2c.

Komzpa:
Rebased onto current upstream/main d156ebd and removed the duplicate base64 dependency exposed by the rebase. New head: afda133. Clean Debian trixie cargo test --lib: 110/110 passed.

PR #272, opened 2026-07-31
Add headless Matrix history paging command

Komzpa:
## Summary

- add `/matrix history` to fetch one older room page without requiring a WeeChat window
- emit a hidden, machine-readable completion line with the added count and exhaustion state
- keep unavailable, busy, and failed outcomes explicit so relay GUIs can stop or retry safely
- bound restored and interactive requests to 25 events per call so relay GUIs can request further pages explicitly

## Why

Relay-native frontends do not own a real WeeChat window, so `/window page_up` cannot drive Matrix pagination. This command gives them a buffer-scoped paging primitive while preserving the existing interactive behavior.

The companion GUI integration is rnocx/WeeChatRS#18.

## Validation

- `WEECHAT_BUNDLED=1 cargo test --locked`: 94 passed
- `rustfmt --edition 2018 --check src/commands/matrix.rs src/room/mod.rs`
- `git diff --check`

No Matrix message was sent during validation.

## Update: empty history pages

- Keep back-pagination available when Matrix returns an empty chunk with a new `end` cursor instead of treating every empty chunk as exhausted.
- Stop cleanly when the server omits the next cursor or repeats the current cursor, including the non-empty repeated-cursor case that could otherwise replay forever.
- Added focused cursor-state regressions. The integrated backend composite containing this change passed all 101 tests, and all four exact-head build-test-release jobs are green.

### DIFF
--- a/src/commands/matrix.rs
+++ b/src/commands/matrix.rs
@@ -18,6 +18,7 @@ use super::{
 use crate::{
     commands::{DevicesCommand, KeysCommand, MediaCommand, VerifyCommand},
     config::ConfigHandle,
+    room::HistoryPageResult,
     MatrixServer, Servers, PLUGIN_NAME,
 };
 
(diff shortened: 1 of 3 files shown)


strk:
Why a draft ?

Komzpa:
Marked ready for review as requested; the branch is clean and its four hosted package checks were green at the recorded head d72f8cd.

Komzpa:
Rebased onto current upstream/main d156ebd. New head: 29d7568. Clean Debian trixie cargo test --lib: 111/111 passed.

PR #273, opened 2026-07-31
Upload native Matrix attachments from relay clients

Komzpa:
## Summary

- add a bounded `/matrix-upload` transaction for transferring arbitrary attachment bytes over the existing authenticated WeeChat relay command path
- bind every transfer to the Matrix room or thread buffer that started it, then use the Matrix SDK's native attachment upload API with the original filename and MIME type
- preserve the thread relation for both remote relay uploads and the existing local `/upload <path>` command
- advertise `matrix_upload_v1=1` on room and thread buffers so relay clients can enable native upload controls only when the backend supports them
- consume handled `/matrix-upload` commands with `OkEat` so normal WeeChat dispatch cannot append a misleading unknown-command error

## Why

The existing upload command accepts only a path on the WeeChat host. A remote GUI therefore cannot send a local screenshot or selected file without either sharing a filesystem or first uploading it to an unrelated external file host. The existing attachment path also drops the initiating thread relation, so an upload started in a thread lands in the main room.

This change keeps Matrix attachments inside the authenticated relay and Matrix connection. Transfers use ordered URL-safe base64 chunks, an immutable transfer ID and buffer target, a declared size, a validated MIME type, and a SHA-256 digest. The receiver rejects malformed, out-of-order, oversized, stale, hash-mismatched, or target-mismatched transfers before calling the Matrix SDK. PNG payloads additionally retain their signature check; other MIME types preserve their exact bytes.

Room and thread buffers expose an explicit `matrix_upload_v1` local variable for capability negotiation. The command hook returns `OkEat` after handling every transaction step, preventing the same input from falling through to normal command dispatch.

The companion GUI integration is in rnocx/WeeChatRS#18.

## Limits and cleanup

- 10 MiB maximum decoded payload
- 32 KiB maximum encoded chunk
- four concurrent transfers
- abandoned transfers expire after 60 seconds

## Validation

- Rust 1.96 Bookworm container with the packaged WeeChat development header
- full unit-test suite, including PNG, PDF, and arbitrary binary byte round-trips; MIME, ordering, size, hash, expiry, room attachment, and thread attachment coverage
- 100 unit tests passed on the current head
- isolated WeeChat plugin smoke confirmed that a handled `/matrix-upload` command is not followed by an unknown-command error
- `rustfmt --check` for all changed Rust files
- `git diff --check`


strk:
conflicts, and why is this a draft ?

Komzpa:
Marked ready for review as requested; the branch is clean and its four hosted package checks were green at the recorded head 65092e2.

Komzpa:
Rebuilt the rebase from exact prior remote head 65092e2 (preserving its repair commits), rebased onto current upstream/main, and removed the duplicate base64 dependency exposed by the rebase. New head: 5deefeb. Clean Debian trixie cargo test --lib: 117/117 passed.

PR #274, opened 2026-08-01
Retry Matrix events when missing room keys arrive

Komzpa:
## Summary

- retain Megolm events that could not initially be decrypted and request their missing sessions from the configured key backup
- re-fetch encrypted events represented by restored WeeChat log placeholders, so a restart does not lose the ciphertext needed for a later retry
- retry decryption after room-key delivery or key import and replace the existing main-room or thread-buffer lines in place

This keeps the normal placeholder when no usable key exists. If the key later becomes available, the existing line is re-rendered as the decrypted event instead of requiring a duplicate timeline event.

## Testing

- `WEECHAT_BUNDLED=1 cargo test --locked --lib` (94 passed)
- `WEECHAT_BUNDLED=0 WEECHAT_PLUGIN_FILE=/usr/include/weechat/weechat-plugin.h cargo build --release --locked --offline`
- loaded the release plugin in a temporary WeeChat 4.9.5 headless profile

### DIFF
--- a/src/room/buffer.rs
+++ b/src/room/buffer.rs
@@ -203,6 +203,45 @@ impl RoomBuffer {
             })
     }
 
+    /// Return encrypted event IDs represented by placeholder lines restored
+    /// from WeeChat logs or currently displayed in a room/thread buffer.
+    pub fn encrypted_event_ids(&self) -> Vec<OwnedEventId> {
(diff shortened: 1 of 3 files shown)


strk:
Trusted merge. I'm thinking these files are getting too big, maybe a refactoring would be due at some point.

PR #278, opened 2026-08-03
Fix /topic runtime panic

### LINKED ISSUE #266: Panic in main Weechat thread when using /topic with a message argument
Running `/topic <msg>` in a room buffer causes a panic:

```
23:39:45 =!= | Panic in the main Weechat thread: panicked at
             | /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/ops/function.rs:250:5:
23:39:45     | there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

The `/topic` command triggers an async operation outside of a Tokio runtime context. Likely the same class of bug as the `/part` panic (#212) and the `/redact` panic (#207) that were fixed by ensuring async work runs inside the runtime.

Komzpa:
Fixes #266.

The /topic command used to call Matrix SDK room.set_room_topic() from the WeeChat executor. That can panic because the SDK HTTP path expects the Matrix Tokio runtime.

This adds a MatrixRoom::set_topic() helper that routes the SDK call through the room connection runtime, matching the existing send/redact runtime boundary.

Local verification:
- rustfmt --edition 2021 --check src/commands/topic.rs src/room/mod.rs
- git diff --check

Note: full cargo check/test is blocked on my local host by openssl-sys rejecting system OpenSSL 4.0.1 before crate compilation.

### DIFF
--- a/src/commands/topic.rs
+++ b/src/commands/topic.rs
@@ -2,7 +2,7 @@ use clap::{App as Argparse, AppSettings as ArgParseSettings, Arg};
 use weechat::{
     buffer::Buffer,
     hooks::{Command, CommandCallback, CommandSettings},
-    Args, Prefix, Weechat,
+    Args, Weechat,
 };
 
 use super::parse_and_run;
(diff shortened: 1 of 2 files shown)


Komzpa:
Pushed `f31b1de125e908e33f13539e94c7554b2ee50740` to fix the CI compile error from the first version.

The branch is still for poljar/weechat-matrix-rs#266: run `/topic <text>` on the Matrix connection runtime instead of the main WeeChat thread. Local validation now passes with Rust 1.96.1 and bundled WeeChat:

```sh
WEECHAT_BUNDLED=1 CARGO_TARGET_DIR=/target cargo check --locked
```

GitHub currently reports the PR mergeable; hosted package jobs are running on the new head.

PR #280, opened 2026-08-05
fix: complete room nicks in thread buffers

### LINKED ISSUE #279: Nick TAB-completion does not work in Matrix thread buffers
## Description

Pressing Tab in the input line of a Matrix **thread** buffer does not complete nicknames of room members. The same Tab completion works fine in regular room buffers (and in IRC channel buffers).

Expected behavior: typing the beginning of a member's display name in a thread buffer and pressing Tab should offer/complete the matching room member nicks, same as it does in the parent room buffer.

## Root cause (found in source)

Thread buffers are created with the nicklist explicitly disabled, in `Room::get_or_create_thread_buffer()` (src/room/mod.rs):

```rust
buffer.set_short_name(&thread_short_name);
buffer.enable_multiline();
buffer.disable_nicklist();          // <-- nicklist disabled
buffer.disable_nicklist_groups();
```

WeeChat's built-in `%(nicks)` completion reads the current buffer's nicklist, so with the nicklist disabled there are no candidates to complete from in thread buffers. Room buffers (which do `enable_nicklist()` + populate members) complete fine.

## Possible directions

- Don't call `disable_nicklist()` on thread buffers, so they inherit/use the room's members for nick completion (note: this would also show the nicklist bar in thread buffers, which may or may not be desired).
- Alternatively hook a custom completion item (e.g. `nicks`) for thread buffers that falls back to the parent room's members, keeping the thread buffer's own nicklist empty/absent.

## Repro

1. Open a thread from a room with several members.
2. In the thread buffer's input line, type the first few letters of a member's display name.
3. Press Tab: nothing is completed.
4. In the parent room buffer, the same gesture completes nicks from the room nicklist.

strk-ai-agent:
Fixes #279

## Problem

TAB-completion of room member nicks does not work in Matrix thread buffers, while it works in regular room buffers.

WeeChat's built-in `nicks` completion (`weechat-completion.c`, `completion_list_add_nicks_cb`) first runs plugin hooks named `nick`; if none add candidates, it falls back to iterating the current buffer's nicklist. Thread buffers are created with an empty nicklist (`Room::get_or_create_thread_buffer` in `src/room/mod.rs`), so there are no candidates to complete from.

## Change

Hook the `nick` completion item in `src/completions.rs`. For thread buffers (detected via the `thread_root` localvar) it adds the parent room's members through the existing `Servers::find_room` / `RoomHandle::names` path, which resolves thread buffers to their parent room.

For all other buffers the hook adds nothing, so WeeChat falls back to the built-in nicklist-based completion, leaving room-buffer behavior unchanged. The IRC plugin's own `nick` hook already coexists with this mechanism (it guards on IRC buffers), so no interference is expected.

## Verification

- `cargo check --all-targets` passes.
- `cargo test` passes (108 tests).
- Behavior is only additive: candidates are contributed for thread buffers, and only via the existing members data.

Not done: thread buffers still keep an empty nicklist (no sidebar), matching current UI; only their completion is fed from the parent room.

### DIFF
--- a/src/completions.rs
+++ b/src/completions.rs
@@ -19,6 +19,7 @@ pub struct Completions {
     servers: CompletionHook,
     users: CompletionHook,
     media: CompletionHook,
+    nicks: CompletionHook,
 }
 
 impl Completions {
@@ -29,7 +30,8 @@ impl Completions {
             )?,
             servers: ServersCompletion::create(servers.clone())?,
             users: UsersCompletion::create(servers.clone())?,
-            media: MediaCompletion::create(servers)?,
+            media: MediaCompletion::create(servers.clone())?,
+            nicks: NicksCompletion::create(servers)?,
         })
     }
 }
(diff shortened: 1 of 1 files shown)


strk:
Please add a test for this new code

strk-ai-agent:
Added a unit test covering the new logic (commit e423eea). Since the completion callback operates on WeeChat's C API (Buffer/Completion), it can't be exercised in unit tests, so the candidate decision is extracted into a pure helper `nick_completion_candidates` and tested: thread buffers contribute the parent room's members sorted, a thread buffer without a room contributes nothing, and non-thread buffers contribute nothing so the built-in nicklist completion is preserved. Full suite: 111 tests pass.

PR #281, opened 2026-08-06
feat: add room access control commands

### LINKED ISSUE #267: Add room access control commands (public, invite-only, etc.)
There is currently no way to change a room's access settings from within Weechat — e.g. making a room public, invite-only, or setting join rules.

It would be useful to have commands like:

- `/room make_public` (set join rule to `public`)
- `/room make_invite_only` (set join rule to `invite`)
- `/room set_join_rule <rule>` (generic)

These would call the Matrix `PUT /_matrix/client/v3/rooms/{roomId}/state/m.room.join_rules` endpoint.

Komzpa:
Adds the room access-control commands requested in #267:

- `/room make_public`
- `/room make_invite_only`
- `/room set_join_rule public|invite|knock|private`

The state event is sent through the existing Matrix connection runtime and reports success or failure in the room buffer. Restricted join rules remain rejected because they require an allow-list rather than a single rule token. Includes parser coverage for all supported rules.

Fixes #267.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -30,12 +30,15 @@ fn main() {
 
     // Capture git describe output for the version command
     if let Ok(output) = std::process::Command::new("git")
-        .args(["describe", "--long", "--tags", "--always", "--all", "--dirty"])
+        .args([
+            "describe", "--long", "--tags", "--always", "--all", "--dirty",
+        ])
         .current_dir(std::env::var("CARGO_MANIFEST_DIR").unwrap())
         .output()
     {
         if output.status.success() {
-            let describe = String::from_utf8_lossy(&output.stdout).trim().to_string();
+            let describe =
+                String::from_utf8_lossy(&output.stdout).trim().to_string();
             println!("cargo::rustc-env=GIT_DESCRIBE={}", describe);
         }
     }
--- a/src/commands/access.rs
+++ b/src/commands/access.rs
@@ -13,6 +13,8 @@ use crate::Servers;
 enum AccessAction {
     Public,
     Invite,
+    Knock,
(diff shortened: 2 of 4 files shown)


Komzpa:
Focused implementation for #267 is pushed at ec22974. It adds `/room make_public`, `/room make_invite_only`, and `/room set_join_rule public|invite`, with parser tests and Matrix runtime error reporting. `git diff --check` passes; local cargo test is blocked by this hosts unsupported OpenSSL 4.0.1, so hosted CI is the remaining build gate.

Komzpa:
Correction with stronger validation: the full `cargo test --lib` now passes in a fresh Debian trixie container (Rust 1.97.1, 111/111 tests). The host-only OpenSSL 4.0.1 limitation does not apply there. GitHub has not started a workflow for this head yet.

Komzpa:
Follow-up head 9db3569 adds the remaining simple Matrix join rules (`knock` and `private`) to the generic command. Fresh Debian trixie `cargo test --lib`: 111/111 passing.

strk:
Is `/room make_public` the same as `/room set_join_rule public` ? 
Is `/room make_invite_only` the same as `/room set_join_rule invite` ?

Why the duplication ? I'd only keep the `set_join_rule` to reduce the maintenance burden

Komzpa:
Yes, they were aliases. I removed the duplicate subcommands and kept only `/room set_join_rule public|invite|knock|private` in 346f35d. Fresh container check: `cargo test --lib` on Rust 1.97/Debian bookworm passed 110/110. GitHub checks are queued.

PR #289, opened 2026-08-11
Retry Matrix network reconnects

Komzpa:
## Summary

Transient network outages could leave the Matrix client requiring a manual
reconnect or resend. The sync loop retried immediately on errors, stopped
forever if `sync_with_callback()` returned cleanly while the WeeChat channel was
still alive, and message sends failed after one transport attempt.

This adds bounded retry/backoff for Matrix send requests using the same
transaction ID, retries sync filter upload after network errors, and restarts an
unexpectedly completed sync loop unless shutdown is already in progress.

## Stack relationship

This PR and poljar/weechat-matrix-rs#290 both change connection lifecycle code;
their exact heads produce a content conflict in `src/connection.rs`. If either
lands first, the other must be rebased while preserving both the reconnect
backoff and the bounded room-upgrade join path.

## Validation

- `cargo test --locked --lib`
- all four exact-head `build-test-release` jobs passed on
  `ffff2d7ba89c81cd8d42c58320f2b21f7c7e48f6`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ dashmap = "6.1.0"
 url = "2.3.1"
 serde_json = "1.0.85"
 strum = { version = "0.24.0", features = ["derive"] }
-tokio = { version = "1.21.1", features = ["rt-multi-thread", "sync"] }
+tokio = { version = "1.21.1", features = ["rt-multi-thread", "sync", "time"] }
 tracing = "0.1.36"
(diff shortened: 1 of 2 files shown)


strk:
is WIP indicator intentional ?

Komzpa:
No, not intentional anymore. This one is a single focused reconnect-retry commit and the current head `ffff2d7` is green on all four package jobs, so I marked it ready.

PR #294, opened 2026-08-30
reply: select events by slash pattern

### LINKED ISSUE #287: Add pattern support to /reply
Like I believe is done with /redact, we could support pattern matching for /reply too, would make it easier to reply

Komzpa:
Fixes poljar/weechat-matrix-rs#287.

This adds the same slash-delimited, case-sensitive substring selection that
`/redact` already has to `/reply`, so `/reply /needle/ message` replies to the
most recent unredacted Matrix event whose rendered line contains `needle`.

## Relationship to thread replies

poljar/weechat-matrix-rs#270 and this PR both change the `/reply` implementation.
Each head is independently mergeable with `main`, but merging the two heads
together produces a content conflict in `src/commands/reply.rs`. If
poljar/weechat-matrix-rs#270 lands first, this change must be rebased onto it so
the pattern lookup retains poljar/weechat-matrix-rs#270's thread relation and
thread-root behavior.

## Validation

- `git diff --check`
- `rustfmt --edition 2018 --check src/commands/reply.rs`
- clean Debian trixie container with bundled WeeChat headers:
  `cargo test --locked --lib` (134 passed)
- [hosted build-test-release run](https://github.com/poljar/weechat-matrix-rs/actions/runs/33339978313):
  three Ubuntu build/test jobs and the Debian 13 package job passed on head
  `82dedee4105a595f7e9f216d27341551e39324da`

### DIFF
--- a/src/commands/reply.rs
+++ b/src/commands/reply.rs
@@ -17,6 +17,10 @@ use crate::Servers;
 
 const EVENT_ID_TAG_PREFIX: &str = "matrix_id_";
 
+fn message_matches_pattern(message: &str, pattern: &str) -> bool {
+    message.contains(pattern)
+}
+
 pub struct ReplyCommand {
     servers: Servers,
 }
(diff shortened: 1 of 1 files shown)


strk:
conflicts

Komzpa:
Rebased onto current upstream/main (`e24322a`). New head: `458a86c`. Local merge-tree against upstream/main is clean (`454e747d4203b896f54d197d546efb7a791556a3`). Validation: `cargo test --lib` passed in the Docker gate with `/usr/include/weechat/weechat-plugin.h` mounted from the live WeeChat 4.10.0 install: 145 passed, 0 failed.

PR #311, opened 2026-09-16
fix: surface actionable guidance for undecryptable messages

### LINKED ISSUE #308: Actionable guidance when a sender's client is not sharing room keys
When a room message cannot be decrypted because the sender's device is not sharing the room key (key withheld, or message encrypted to a session we don't hold), the plugin currently shows only a generic placeholder in the room buffer:

    <Unable to decrypt message>

(see `Render for RoomEncryptedEventContent` in src/render.rs; it even has a TODO to add the sender key / algorithm / session id to the tags).

In practice the resolution requires the sender to re-send the message or forward the room key from one of their other devices, and there is no way to tell from the placeholder what went wrong or what to do about it.

Proposed improvements:
1. Surface the specific failure reason from the SDK (e.g. key withheld vs unknown inbound session id, which sender device is involved) — either in the placeholder or as a hint line.
2. Automatically send a room-key request to the sender when decryption fails for a missing session. #274 (closed) already retains undecryptable events and retries once keys arrive or are imported, but it does not actively request keys from the sender.
3. Consider a per-room command to re-request the room key / show the "re-send message or forward the key" guidance inline (also useful on the mobile-web/E2EE UX parity front).

strk-ai-agent:
Fixes #308

When a room event cannot be decrypted because the sender's client is not sharing the room key, the plugin currently shows only a bare `<Unable to decrypt message>` placeholder. This makes the failure reason and recovery steps visible:

1. The undecryptable placeholder now names the sender and adds an actionable hint line: ask them to re-send the message or forward the room key from another device, noting the message will be decrypted automatically once the key arrives (the retry machinery from #274 applies).

2. `m.room_key.withheld` to-device events are now handled. The sender's client tells us why the key was not shared (unverified device, blacklisted, unauthorised, unavailable, no Olm session), including the sender's own reason string, printed with a Network prefix to the room buffer when the room is known.

No new config options; the hints are unconditional like the previous placeholder.

Verification: `cargo test --lib` (170 passed), `cargo check` clean, no new clippy warnings.


strk:
@Komzpa do you want to ask your Minion to review this ?

src/server.rs
```diff
@@ -1668,6 +1668,73 @@ impl InnerServer {
...
+                    RoomKeyWithheldCodeInfo::Unauthorized(_) => {
+                        "your device is not allowed to receive the key"
+                    }
+                    RoomKeyWithheldCodeInfo::NoOlm => {
+                        "no Olm session could be established with your device"
+                    }
+                    _ => "the room key was withheld",
+                };
+
+                let mut message = format!(
```
Komzpa: `m.unverified`, `m.blacklisted`, and `m.no_olm` are terminal conditions for the current key request, but this always tells the user to ask the sender to re-send or forward the key. Re-sending will normally produce the same withheld response. Give each recognized code its own remediation (verify the device, remove the blacklist, or establish an Olm session); only `m.unavailable` can usefully suggest another device.

src/server.rs
```diff
@@ -1668,6 +1668,73 @@ impl InnerServer {
...
+                    "{} withheld the room key: {}. Ask them to re-send their \
+                     message or forward the room key from another of their \
+                     devices.",
+                    e.sender, why
+                );
+                if let Some(reason) = &e.content.reason {
```
Komzpa: Ruma documents `reason` as a human-readable fallback to use only when the receiving client does not understand `code`; this handler has already recognized the code. Appending arbitrary `reason` here can contradict the normalized guidance. Please reserve it for the unknown-code branch (ideally including that unknown code) and omit it for known codes.

Komzpa (changes requested):
Two changes are needed so the promised recovery guidance stays accurate.

src/server.rs
strk-ai-agent: @Komzpa good catch, thanks. The handler now gives each recognized code its own remediation: m.unavailable suggests re-sending or forwarding from another device; m.unverified says to verify the device (the plugin has verification support); m.blacklisted says to unblacklist; m.unauthorized says to request the key; m.no_olm says to establish an Olm session (both devices online). Re-sending is only suggested for m.unavailable, as you noted.

src/server.rs
strk-ai-agent: @Komzpa agreed — the raw `reason` is now only appended in the fallback branch for unrecognized codes, together with the code itself (e.g. `withheld the room key with code: X. Reason given: ...`). Known codes omit it entirely.

src/render.rs
```diff
@@ -629,25 +629,49 @@ impl<C: HasUrlOrFile> Render for C {
...
+            // in time, sender key, algorithm, session id.
+            tags: tags.clone(),
+        }];
+
+        // The room key is missing, which usually means the sender's client is
+        // not sharing keys with us. Print actionable guidance: only the sender
+        // can help (re-send the message or forward the room key from another
+        // device). The event is retried and re-rendered automatically once the
+        // key arrives (see #274).
+        lines.push(RenderedLine {
```
Komzpa, line 659: This makes every undecryptable placeholder two lines. When the key arrives, `replace_event_helper` in src/room/buffer.rs rewrites the placeholder lines in place and, because WeeChat cannot delete a line, blanks any surplus. A placeholder that decrypts to the usual one-line message therefore leaves a permanent empty line in the buffer, once per recovered message. It also doubles the line count of an undecryptable backlog on history backfill.

Please keep the placeholder at one line (append the hint to the same message) so the later replacement stays line-for-line.

src/server.rs
```diff
@@ -1668,6 +1668,95 @@ impl InnerServer {
...
+                        }
+                        RoomKeyWithheldCodeInfo::Unavailable(data) => {
+                            Some(data)
+                        }
+                        RoomKeyWithheldCodeInfo::NoOlm => None,
+                        _ => None,
+                    };
+
+                let message = match code {
+                    RoomKeyWithheldCodeInfo::Unavailable(_) => format!(
```
Komzpa, line 1700: With the SDK's default `automatic-room-key-forwarding` feature, the key requests that trigger `m.unavailable` / `m.unauthorised` replies are sent only to our own devices (`GossipRequest.request_recipient` is the own user id in matrix-sdk-crypto 0.18; nothing in this plugin requests keys from the original sender). So in the common case `e.sender` here is the user's own MXID and this prints `@me:example.org withheld the room key: the room key is not available on their device. Ask them to re-send their message ...`.

Please branch on `e.sender` being our own user (`sender_key` identifies which of our devices answered) and say something like "your other device does not have this key"; keep the "ask the sender" wording for a foreign sender only.

src/server.rs
```diff
@@ -1668,6 +1668,95 @@ impl InnerServer {
...
+                         key gets shared.",
+                        e.sender
+                    ),
+                    RoomKeyWithheldCodeInfo::Blacklisted(_) => format!(
+                        "{} withheld the room key: your device is \
+                         blacklisted. Ask them to unblacklist your device to \
+                         receive the room key.",
+                        e.sender
+                    ),
+                    RoomKeyWithheldCodeInfo::Unauthorized(_) => format!(
```
Komzpa, line 1720: `m.unauthorised` is a deliberate refusal (spec example: the requesting user/device was not in the room when the message was sent). "Ask them to share the room key with you" is the same kind of advice the previous round removed: the request was already made and answered. Better to state that this device will not be given the key by that sender, and when the sender is our own user, that this device is not entitled to it.

src/server.rs
```diff
@@ -1668,6 +1668,95 @@ impl InnerServer {
...
+                        "{} withheld the room key: your device is not \
+                         allowed to receive it. Ask them to share the room \
+                         key with you.",
+                        e.sender
+                    ),
+                    RoomKeyWithheldCodeInfo::NoOlm => format!(
```
Komzpa, line 1726: `m.no_olm` carries no room or session and applies to every message the sender sent before we receive it (spec: "The recipient may assume that this m.room_key.withheld message applies to all encrypted room messages sent before it receives the message"). There is no single room key to "share again", and the remedy is not both devices being online: the sender needs this device's keys from the homeserver, which this client uploads while it syncs, after which a session can be established and new messages decrypt.

Suggested text: "could not establish an encrypted session with this device; messages they sent before now cannot be recovered from them. Keep this client connected; new messages should decrypt once a session is established."

Komzpa (changes requested):
Thanks, both earlier threads are addressed on 17976c2 and can be resolved. Reviewing the rest of this head, two things still produce wrong output for users and two of the withheld texts contradict what the code means; details inline.

Non-blocking, for a follow-up: matrix-sdk already persists `m.room_key.withheld` info and reports it at decryption time as `UnableToDecryptReason::MissingMegolmSession { withheld_code }` on `TimelineEventKind::UnableToDecrypt`, and it also distinguishes `UnknownMegolmMessageIndex` and malformed events, where "waiting for the room key" is not accurate. Deriving the placeholder hint from that would cover withheld events that arrived while this client was not running, instead of a one-shot Network line.

Komzpa:
I could not update the source branch directly, so I pushed a tested maintainer patch here:

- Branch: https://github.com/Komzpa/weechat-matrix-rs/tree/fix/pr311-undecryptable-guidance-komzpa
- Commit: `030d891` (`fix: refine undecryptable key guidance`)

It addresses the remaining review points on this head:

- keeps the undecryptable placeholder to one rendered line, so later in-place decrypted replacement does not leave a blank surplus line;
- distinguishes own-device `m.unavailable` from a foreign sender;
- changes `m.unauthorised` away from “ask them to share the key” wording;
- changes `m.no_olm` guidance to session-establishment/new-message wording rather than a specific room-key resend.

Validation: `RUSTC=/usr/bin/rustc /usr/bin/cargo test --quiet --lib` passed (`174` tests).

strk-ai-agent:
Thanks @Komzpa! Merged your patch (030d891) into the PR branch, on top of 17976c2. Verified with your exact command: cargo test --lib passes (174 tests). Appreciate the thorough review and the tested maintainer patch.

